PR #439, opened 2021-11-23
Decoder implementation for Recognizers.

horned-sphere:


SirCipher (approved):

api/formats/swim_recon/src/parser/async_parser/mod.rs
```diff
@@ -351,3 +353,195 @@ impl LocationTracker {
         })
     }
 }
+
+/// Tokio [`Decoder`] that parses its input as Recon and uses this to drive a [`Recognizer`] until
+/// it completes. Not that this is cannot be used as a stand-alone decoder as it has no concept of
+/// a separator between frames. I needs to be incorporated into another decoder that can determine
```
DobromirM, line 359:
```suggestion
/// a separator between frames. It needs to be incorporated into another decoder that can determine
```

api/formats/swim_recon/src/parser/async_parser/mod.rs
```diff
@@ -351,3 +353,195 @@ impl LocationTracker {
...
+            location,
+        } = self;
+
+        let mut current = span;
+        let result = 'outer: loop {
+            match parser.parse(current) {
+                Ok((rem, events)) => {
+                    current = rem;
+                    for event_or_end in events {
+                        if let Some(event) = event_or_end {
```
DobromirM, line 401: Could probably use `while let Some(...) `

api/formats/swim_recon/src/parser/async_parser/mod.rs
```diff
@@ -351,3 +353,195 @@ impl LocationTracker {
...
+    span: Span<'a>,
+    location: &mut LocationTracker,
+) -> Result<(Span<'a>, Option<R::Target>), AsyncParseError>
+where
+    R: Recognizer,
+{
+    match parser.parse(span) {
+        Ok((rem, events)) => {
+            for event_or_end in events {
+                if let Some(event) = event_or_end {
```
DobromirM, line 451: Same here.

api/formats/swim_recon/src/parser/async_parser/mod.rs
```diff
@@ -351,3 +353,195 @@ impl LocationTracker {
...
+                                Ok((final_rem, None)) => (final_rem, None),
+                                Ok((final_rem, ow)) => (final_rem, Some(Ok(ow))),
+                                Err(e) => (span, Some(Err(e))),
+                            }
+                        }
+                        _ => (span, None),
+                    };
+                    let finalized = if let Some(res) = final_result {
+                        res
+                    } else {
```
DobromirM, line 521: Maybe this can be done with `unwrap_or_else(...)`

DobromirM (approved):

api/formats/swim_recon/src/parser/async_parser/mod.rs
horned-sphere: Not very nicely. It's an iterator of `Option`s and `None` and `Some<None>` from the iterator get treated differently. Converting the loop to `while let...` doesn't really make it any better.

horned-sphere: You end up with:
```
let mut it = events.into_iter();
while let Some(event_or_end) = it.next() {
    if let Some(event) = event_or_end {
          ...
    } else {
        break ...;
    }
}
```
which I don't think is much of an improvement.

PR #451, opened 2021-11-30
Raw envelope header

SirCipher:


api/formats/swim_recon/src/parser/header.rs
```diff
@@ -0,0 +1,332 @@
...
+            },
+            HeaderParserState::Fin => Ok((input, ReadEvent::End)),
+        }
+    }
+}
+
+pub struct HeaderParseIterator<'a> {
+    input: Span<'a>,
+    parser: Option<HeaderParser>,
+}
```
horned-sphere, line 102: I don't think there's a need for the header parser to be stateful and resumable. The header will generally be a very short string. I think it would be better to just have a very basic non parser, attempt it against the buffer and, if it's incomplete, just run it again. (I'm not entirely sure that it wouldn't be better to do that for _everything_ but I'm trying to the philosophy of the Java version).

SirCipher: That's understandable. I can rip the iterator out and just expose the raw nom function.

PR #452, opened 2021-12-08
Newtype derive

DobromirM:


api/swim_form/src/structural/read/recognizer/mod.rs
```diff
@@ -1508,6 +1508,104 @@ impl<T, Flds> Recognizer for OrdinalStructRecognizer<T, Flds> {
...
+
+impl<T, Flds> LabelledNewtypeRecognizer<T, Flds> {
+    /// # Arguments
+    /// * `fields` - The state of the recognizer state machine (specified by the macro).
+    /// * `vtable` - Functions that are generated by the macro that determine how incoming events
+    /// modify the state.
+    pub fn new(
+        _tag: TagSpec,
+        fields: Flds,
+        _num_fields: u32,
```
SirCipher, line 1527: Unused fields

api/swim_form/src/tests/derive/form/newtype_empty_struct.stderr
```diff
@@ -0,0 +1,5 @@
+error: Cannot apply `newtype` attribute to an empty struct
+ --> src/tests/derive/form/newtype_empty_struct.rs:4:14
+  |
+4 |     #[derive(Form)]
+  |              ^^^^
```
SirCipher, line 5: Is it possible to change the error span to be the struct itself?

horned-sphere: Recently, the compiler seems to have been ignoring the span that you tell it. If you look at the expectations in the existing Form derive tests, they've all ended up like this. I don't know if that is a deliberate change or a bug but for now, targetting the error doesn't seem to work.

api/swim_form/src/structural/read/recognizer/mod.rs
horned-sphere: I think this might be what @DobromirM  meant by reusing some of the macro code by leaving on extra parameters. Now that I see it, I think it might be better to have more code in the macros to avoid it.

api/swim_form_derive/src/structural/model/record/mod.rs
```diff
@@ -28,7 +28,12 @@ use std::ops::Add;
 use swim_utilities::errors::validation::{validate2, Validation, ValidationItExt};
 use syn::{Attribute, Fields, Ident};
 
-/// Description of the fields, taken from the derive input, preprocessed with any modifcations
+const NEWTYPE_MULTI_FIELD_ERR: &str = "Cannot apply `newtype` attribute to a multi-field struct";
```
horned-sphere, line 31: "struct with multiple fields" sounds better

horned-sphere (approved):

SirCipher (approved):

PR #453, opened 2021-12-10
Fixes RoutingAddr creation

SirCipher:
Also updates the display implementation to display an integer value.

### DIFF
--- a/runtime/swim_runtime/src/routing/mod.rs
+++ b/runtime/swim_runtime/src/routing/mod.rs
@@ -15,6 +15,7 @@
 use crate::error::{ConnectionDropped, ResolutionError, RouterError, RoutingError};
 use std::convert::TryFrom;
 
+use bytes::Buf;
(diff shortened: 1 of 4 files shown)


DobromirM (approved):

runtime/swim_runtime/src/routing/mod.rs
```diff
@@ -82,15 +82,21 @@ impl RoutingAddr {
     pub fn uuid(&self) -> &Uuid {
         &self.0
     }
+
+    fn get_location(&self) -> u32 {
+        let RoutingAddr(location) = self;
+        let repr = <[u8; 4]>::try_from(&location.as_bytes()[12..]).unwrap();
+        u32::from_be_bytes(repr)
```
horned-sphere, line 89: You could just use the `Buf` trait from the `bytes` crate and do get_u32 on the slice.

horned-sphere (approved):

PR #456, opened 2022-01-07
Incremental recon comparison

DobromirM:


api/formats/swim_recon/src/comparator/mod.rs
```diff
@@ -0,0 +1,306 @@
...
+    NoKey,
+    Attr,
+    Slot,
+}
+
+#[derive(Debug, Clone)]
+struct BuilderState {
+    key: KeyState,
+    in_body: bool,
+}
```
horned-sphere, line 227: These could be Copy.

api/formats/swim_recon/src/comparator/mod.rs
```diff
@@ -0,0 +1,306 @@
...
+#[derive(Debug, Clone)]
+enum ValidatorState {
+    Init,
+    InProgress(InProgressState),
+    Invalid,
+}
+
+#[derive(Debug, Clone)]
+struct InProgressState {
+    stack: Vec<BuilderState>,
```
horned-sphere, line 100: This could be a good place to use the `smallvec` crate (or something similar) with a size of about 4. That way, for a typical recon value it wouldn't need to allocate at all.

horned-sphere (commented):
Looks good in general. I've just added a couple of suggestions. (Some deeper test cases might be a good idea).

api/formats/swim_recon/src/comparator/tests.rs
```diff
@@ -0,0 +1,232 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::comparator::{compare_values, incremental_compare};
+use crate::parser::{ParseError, ParseIterator, Span};
+use std::borrow::Cow;
+use swim_form::structural::read::event::{NumericValue, ReadEvent};
+use swim_model::Value;
+
```
horned-sphere, line 20: You should add a test case with deeper nesting to make sure the logic of your stack is correct.

SirCipher (commented):
Looks good but it could do with some documentation; what the boolean arguments denote and what the results mean

SirCipher (approved):

horned-sphere (approved):

PR #465, opened 2022-01-20
Recon comparison improvements

DobromirM:
Improvements to the incremental recon comparison.


api/formats/swim_recon/src/comparator/mod.rs
```diff
@@ -130,224 +126,400 @@ fn incremental_compare<
...
-struct InProgressState {
-    /// The internal stack of builders.
-    stack: SmallVec<[BuilderState; 4]>,
-    /// Flag indicating if there is a slot key.
-    slot_key: bool,
+#[derive(Debug, Clone, Eq, PartialEq)]
+enum ItemType {
+    ValueItem(ValueType),
+    Slot(ValueType, ValueType),
 }
```
horned-sphere, line 137: This is a lot less efficeint that the previous version. What is the edge case that requires this representation?

DobromirM: This was done because the `pop` method of the comparator needs to distinguish between single values and multiple values / slots for cases like this:

```
@foo(@bar@baz)
@foo({@bar@baz})
```

However, I think I might be able to optimise it  because the current version does not compare the items directly.

api/formats/swim_recon/src/comparator/mod.rs
```diff
@@ -127,227 +127,443 @@ fn incremental_compare<
...
     Invalid,
 }
 
-#[derive(Debug, Clone)]
-struct InProgressState {
-    /// The internal stack of builders.
-    stack: SmallVec<[BuilderState; 4]>,
-    /// Flag indicating if there is a slot key.
-    slot_key: bool,
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
```
horned-sphere, line 137: Maybe add some comments to explain what the parts are for (and how they deal with the edge cases).

horned-sphere (approved):

SirCipher (approved):

PR #467, opened 2022-01-25
Travis CI migration.

DobromirM:


horned-sphere (approved):

azure-pipelines.yml
```diff
@@ -3,11 +3,22 @@ pr: ["main"]
...
-  - template: ci/azure-test-stable.yml
-    parameters:
-      name: test_crates
-      rust: 1.56.1
-      displayName: Test all crates
-      cross: eq(variables['Build.SourceBranch'], 'refs/heads/main')
+stages:
+  - stage: Lint
+    jobs:
+      - template: ci/azure-lint.yml
```
SirCipher, line 11: As this is the first step it would be worth moving the cache initialisation to here/before here.

SirCipher (changes requested):

azure-pipelines.yml
DobromirM: I tried it but was getting the following error: `"error: Unrecognized option: \'E\'\n\n"` which turns out is related to using clippy with `sccache`:

https://github.com/mozilla/sccache/issues/1000
https://github.com/rust-lang/rust-clippy/issues/3840

SirCipher (approved):

PR #469, opened 2022-01-26
Recon hash

DobromirM:


api/formats/swim_recon/src/hasher/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+use std::error::Error;
+use std::fmt::{Display, Formatter};
+use std::hash::{Hash, Hasher};
+use swim_form::structural::read::event::ReadEvent;
+
+#[cfg(test)]
+mod tests;
+
+pub fn calculate_hash(value: &str) -> Result<u64, HashError> {
+    let mut hasher = DefaultHasher::new();
```
horned-sphere, line 13: This should probably allow different hashers.

api/formats/swim_recon/src/hasher/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+            if attr_contents.is_implicit_record() {
+                stack.push(ReadEvent::StartBody)?;
+                stack.extend(attr_contents)?;
+                stack.push(ReadEvent::EndRecord)?;
+            } else {
+                stack.extend(attr_contents)?;
+            }
+
+            stack.push(event)?;
+        } else {
```
horned-sphere, line 47: This should just be a match statement.

api/formats/swim_recon/src/hasher/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+    contents: smallvec::SmallVec<[ReadEvent<'a>; 10]>,
+    /// Shows how deep we are in nested records.
+    nested_level: usize,
+    /// Flag showing if the attribute contains any slots.
+    has_slot: bool,
+    /// The number of items that the attribute contains.
+    items: usize,
+    /// Flag showing whether or not the last item that was added was an attribute.
+    is_last_attr: bool,
+}
```
horned-sphere, line 110: Rather than storing the content of the attribute, why not just compute two different hashes and then chose the approprite one when it becomes unambiguous?

api/formats/swim_recon/src/hasher/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+
+pub fn calculate_hash(value: &str) -> Result<u64, HashError> {
+    let mut hasher = DefaultHasher::new();
+
+    let stack = normalise(&mut crate::parser::ParseIterator::new(
+        Span::new(value),
+        false,
+    ))?;
+
+    stack.hash(&mut hasher);
```
horned-sphere, line 20: It seems inefficeint to build up the huge stack of values then do the hash at the end. It would be better to do the has as you go along and just use the stack to keep track of ambiguity.

SirCipher (approved):

PR #477, opened 2022-02-16
Mailbox

horned-sphere:


swim_utilities/multi_reader/src/mailbox/core.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+}
+
+impl<T> Drop for Queue<T> {
+    fn drop(&mut self) {
+        unsafe {
+            let mut cur = *self.tail.get();
+            while !cur.is_null() {
+                let next = (*cur).next.load(Ordering::Relaxed);
+                drop(Box::from_raw(cur));
```
horned-sphere, line 262: Should there be some sort of destructor guard here in case `T`s descructor panics?

swim_utilities/multi_reader/src/mailbox/fan_in/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+pub struct ChannelReceiver {
+    shared: Arc<SharedInner>,
+    queue: QueueConsumer<WriteTask>,
+}
+
+impl Drop for ChannelReceiver {
+    fn drop(&mut self) {
+        self.shared.closed.store(true, Ordering::Relaxed);
```
horned-sphere, line 139: I think the closed flag should be done with Acquire/Release.

swim_utilities/multi_reader/src/mailbox/fan_in/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+                    _ => break,
+                }
+            }
+
+            if queue.has_next() {
+                // Entries inserted into the queue won't wake us up if there is another
+                // entry in the queue. Instead, we register ourselves to be woken up
+                // once we know we are done with the head and that there is another
+                // entry available.
+                cx.waker().wake_by_ref();
```
horned-sphere, line 223: In effect, this means we yield to the runtime ever time we consume from the queue. Might it be better to allow it to conusume several things from the queue before yielding?

swim_utilities/multi_reader/src/mailbox/fan_in/send.rs
```diff
@@ -0,0 +1,128 @@
...
+use std::task::{Context, Poll};
+use tokio_util::codec::Encoder;
+
+pin_project! {
+    pub struct Send<'r,  I> {
+        tx: &'r RawChannelSender,
+        item: I,
+        state: SendState,
+        write_idx: usize
+    }
```
horned-sphere, line 33: Nothing in there is pinned so I don't think the `pin_project` is required. Also, instead of keeping the `write_index` separately, the `item` could be updated each time data is consumed to move its intrinsic start (this is implicitly storing the offset twice).

swim_utilities/multi_reader/src/mailbox/fan_in/send.rs
```diff
@@ -0,0 +1,128 @@
...
+                    let task = WriteTask::new(cx.waker());
+                    let wake = !write_queue.has_next();
+
+                    write_queue.push(task.clone());
+                    *this.state = SendState::Waiting(task);
+
+                    if wake {
+                        // We only want to wake the receiver if there is not another entry in the
+                        // queue as nothing will happen to the task until it has reached the head.
+                        receiver.wake();
```
horned-sphere, line 88: If the queue becomes empty between line 80 and here, I think we'll miss a wakeup that is required?

swim_utilities/multi_reader/src/mailbox/fan_in/send.rs
```diff
@@ -0,0 +1,128 @@
...
+                    // guard against spurious wake ups
+                    if task.is_pending() {
+                        task.register(cx.waker());
+                        return Poll::Pending;
+                    }
+
+                    *this.state = SendState::Writing(task.clone());
+                }
+                SendState::Writing(task) => {
+                    let data = &mut *mailbox.lock();
```
horned-sphere, line 103: I think it should be possible to get rid of this mutex (but it would probably be nasty and wouldn't necessarily make it faster).

horned-sphere (commented):
This is very nice but not quite what I had in mind at first for this component. It's a special purpose channel rather than being someting that could be use to tie other IO producers/consumers together. Given the performance improvment it's probabl worth going in this direction anyway. I wonder if it would be possible to unify the writers/readers of this with those for ByteChannel (so that they could be the same type).

swim_utilities/multi_reader/src/mailbox/fan_in/mod.rs
SirCipher: The idea of this and the https://github.com/swimos/swim-rust/pull/477/files/f4ca609c3f7f23d73a8d5c10bd935238c68c0e54#diff-3d9c430f57d6c5b6ceed898aba63b43f53288dcc2c097448bedea934d6a3695aR88 is that if the queue has entries then the receiver has control over itself. As in, if a producer wants to write and there are entries then don't wake up wake it as it will wake itself here. As you've pointed out, that may result in a race condition between the receiver consuming an entry and the producer not waking it. It's probably best to benchmark it with and without the logic and if it is faster with it then to refactor it into something more robust.

PR #480, opened 2022-02-22
Rust implementation of value and event downlinks.

horned-sphere:
This is actually a lot smaller than it looks. Most of the lines of code are just the builders for the downlink lifecycles.


swim_downlink/src/model/lifecycle/mod.rs
```diff
@@ -0,0 +1,1090 @@
...
+            on_event: self.on_event,
+            on_set: self.on_set,
+            on_unlinked: self.on_unlinked,
+        }
+    }
+
+    /// Replace the handler that is called when the downlink connects with the specified synchronous
+    /// closure. This operation should be guaranteed to complete quickly to avoid blocking the task
+    /// executing the downlink.
+    pub fn on_linked_blocking<F>(
```
SirCipher, line 108: The documentation vs the name of the function is a bit confusing. To me, it implies that this function is allowed to block but the documentation says otherwise?

swim_downlink/src/task/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl<T, LC> Downlink for DownlinkTask<EventDownlinkModel<T, LC>>
+where
+    T: Form + Send + Sync + 'static,
+    <T as RecognizerReadable>::Rec: Send,
+    LC: EventDownlinkLifecycle<T> + 'static,
+{
+    fn kind(&self) -> DownlinkKind {
+        DownlinkKind::Event
+    }
```
SirCipher, line 73: Given that this is essentially a constant, would it be possible to change it to an associated constant?

SirCipher (commented):
Very nice :)

swim_downlink/src/model/lifecycle/mod.rs
```diff
@@ -0,0 +1,1090 @@
...
+mod on_synced;
+mod on_unlinked;
+
+/// The set of handlers that a value downlink lifecycle supports.
+pub trait ValueDownlinkHandlers<'a, T>:
+    OnLinked<'a> + OnSynced<'a, T> + OnEvent<'a, T> + OnSet<'a, T> + OnUnlinked<'a>
+{
+}
+
+/// The set of handlers that a value downlink lifecycle supports.
```
DobromirM, line 36:
```suggestion
/// The set of handlers that an event downlink lifecycle supports.
```

swim_downlink/src/model/lifecycle/mod.rs
```diff
@@ -0,0 +1,1090 @@
...
+            _value_type: PhantomData,
+            on_linked: BlockingHandler(f),
+            on_synced: self.on_synced,
+            on_event: self.on_event,
+            on_set: self.on_set,
+            on_unlinked: self.on_unlinked,
+        }
+    }
+
+    /// Replace the handler that is called when the synchronizes connects.
```
DobromirM, line 125:
```suggestion
    /// Replace the handler that is called when the downlink synchronizes.
```

swim_downlink/src/task/event.rs
```diff
@@ -0,0 +1,122 @@
...
+use tracing_futures::Instrument;
+
+use crate::EventDownlinkModel;
+
+/// Task to drive an event downlink, calling lifecyle events at appropriate points.
+///
+/// #Arguments
+///
+/// * `model` - The downlink model, providing the lifecycle and a stream of values to set.
+/// * `_path` - The path of the lane to which the downlink is attached.
```
DobromirM, line 37:
```suggestion
/// * `path` - The path of the lane to which the downlink is attached.
```

swim_downlink/src/task/event.rs
```diff
@@ -0,0 +1,122 @@
...
+/// * `config` - Configuration parameters to the downlink.
+/// * `input` - Input stream for messages to the downlink from the runtime.
+/// * `_output` - Output stream for messages from the downlink to the runtime.
+pub async fn event_dowinlink_task<T, LC>(
+    model: EventDownlinkModel<T, LC>,
+    path: Path,
+    config: DownlinkConfig,
+    input: ByteReader,
+    _output: ByteWriter,
```
DobromirM, line 46: Will this be removed/used in the future?

swim_downlink/src/task/value.rs
```diff
@@ -0,0 +1,200 @@
...
+use tracing_futures::Instrument;
+
+use crate::ValueDownlinkModel;
+
+/// Task to drive a value downlink, calling lifecyle events at appropriate points.
+///
+/// #Arguments
+///
+/// * `model` - The downlink model, providing the lifecycle and a stream of values to set.
+/// * `_path` - The path of the lane to which the downlink is attached.
```
DobromirM, line 45:
```suggestion
/// * `path` - The path of the lane to which the downlink is attached.
```

DobromirM (approved):

swim_downlink/src/model/lifecycle/mod.rs
horned-sphere: Fair, it's not easy to word it :)

swim_downlink/src/model/lifecycle/mod.rs
```diff
@@ -0,0 +1,1090 @@
...
+        shared_state: Shared,
+    ) -> StatefulValueDownlinkLifecycle<
+        T,
+        Shared,
+        WithShared<FLinked>,
+        WithShared<FSynced>,
+        WithShared<FEv>,
+        WithShared<FSet>,
+        WithShared<FUnlinked>,
+    > {
```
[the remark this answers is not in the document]
horned-sphere: Would it not just say that the typedef is too complex?

swim_downlink/src/task/mod.rs
horned-sphere: It might not be constatnt for all implementations. For example, the JNI implementation of it might need to query something to find out. If, when we're done, all the implementations are constant, we could change it then.

swim_downlink/src/task/event.rs
horned-sphere: The runtime component that this connects to is generic (it's the same for value and event downlinks). It expects there to be a channel in both directions. In this case the `output` isn't needed but it needs to be kept from being dropped to stop the runtime component from shutting down.

swim_downlink/src/model/lifecycle/mod.rs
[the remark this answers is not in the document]
SirCipher: If I'm not mistaken, the rule doesn't run on typedefs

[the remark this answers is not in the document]
horned-sphere: It worked :)

swim_downlink/src/task/value.rs
```diff
@@ -0,0 +1,200 @@
...
+        set_value,
+        lifecycle,
+    } = model;
+    let (stop_tx, stop_rx) = trigger::trigger();
+    let read = async move {
+        let result = read_task(config, input, lifecycle).await;
+        let _ = stop_tx.trigger();
+        result
+    }
+    .instrument(info_span!("Downlink read task.", %path));
```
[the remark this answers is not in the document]
horned-sphere: I've added it at the next level up.

SirCipher (approved):

PR #487, opened 2022-03-11
Relaxed the `Sized` constraint on `StructuralWritable` trait.

DobromirM:


api/swim_form/src/structural/write/mod.rs
```diff
@@ -361,6 +367,23 @@ pub trait BodyWriter: Sized {
     fn done(self) -> Result<Self::Repr, Self::Error>;
 }
 
+impl<T> StructuralWritable for &T
```
horned-sphere, line 370: Add one for `&mut T` too.

SirCipher (approved):

horned-sphere (approved):

PR #483, opened 2022-02-28
Queue for backpressure relief for map downlinks.

horned-sphere:

PR #489, opened 2022-03-14
Runtime component for Map type downlinks.

horned-sphere:
Follows on from #483.

### DIFF
--- a/runtime/swim_runtime/src/downlink/failure.rs
+++ b/runtime/swim_runtime/src/downlink/failure.rs
@@ -20,7 +20,7 @@ use tracing::warn;
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum BadFrameResponse<E> {
-    /// Instruction ignore the bad envelope.
+    /// Instruction to ignore the bad envelope.
     Ignore,
     /// Instruction to abort the downlink runtime task with the specified error.
     Abort(E),
--- a/runtime/swim_runtime/src/downlink/failure.rs
+++ b/runtime/swim_runtime/src/downlink/failure.rs
@@ -1,0 +1,119 @@
+// Copyright 2015-2021 Swim Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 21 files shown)


runtime/swim_runtime/src/downlink/failure.rs
```diff
@@ -0,0 +1,200 @@
...
+
+use std::{convert::Infallible, fmt::Display, num::NonZeroUsize};
+
+use swim_utilities::format::comma_sep;
+use thiserror::Error;
+use tracing::warn;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum BadFrameResponse<E> {
+    /// Instruction ignore the bad envelope.
```
DobromirM, line 23:
```suggestion
    /// Instruction to ignore the bad envelope.
```

DobromirM (approved):

runtime/swim_runtime/src/downlink/interpretation/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+use swim_api::protocol::map::{
+    extract_header, MapMessageEncoder, RawMapOperation, RawMapOperationEncoder,
+};
+use swim_recon::parser::MessageExtractError;
+use tokio_util::codec::Encoder;
+
+#[cfg(test)]
+mod tests;
+
+/// A possible transformation to apply to an incoming event body, before passing it on
```
SirCipher, line 27: Why possibly? Is the transformation only sometimes applied?

runtime/swim_runtime/src/downlink/interpretation/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+/// to the downlink implementation.
+pub trait DownlinkInterpretation {
+    type Error;
+
+    const SINGLE_FRAME_STATE: bool = true;
```
SirCipher, line 32: What is this constant?

runtime/swim_runtime/src/downlink/interpretation/mod.rs
horned-sphere: In the case of a value downlink, it just gets passed through as is. For map downlinks, it gets decomposed into the key and value components first.

runtime/swim_runtime/src/downlink/interpretation/mod.rs
horned-sphere: It dermines of a single frame can provide the entire state for the downlink (value like) or whether multiple frames might be required (map like). I'll add some documentation.

SirCipher (approved):

PR #493, opened 2022-03-17
Recon header peeler.

horned-sphere:
Adds a new capability to the parser for peeling off and interpreting the first attribute of a recon document without creating a `Value` object. Provides an implementation for warp map operations.

### DIFF
--- a/api/formats/swim_recon/src/parser/record/matcher/mod.rs
+++ b/api/formats/swim_recon/src/parser/record/matcher/mod.rs
@@ -17,7 +17,6 @@ use std::str::Utf8Error;
 
 use super::super::tokens::complete::{blob, identifier, numeric_literal};
 use super::super::tokens::string_literal;
-use bytes::Bytes;
 use nom::branch::alt;
 use nom::character::complete::{char, multispace0, one_of, space0};
 use nom::combinator::{complete, flat_map, map, map_res, opt, recognize, rest, success};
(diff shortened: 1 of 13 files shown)


api/formats/swim_recon/src/parser/record/matcher/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    ///
+    /// #Arguments
+    /// * `body` - The remainder of the input span. Note that this is entirely uninterpreted and
+    /// so many not contain valid Recon.
+    fn done(self, body: Span<'a>) -> Result<Self::Output, Self::Error>;
+}
+
+/// Run an implementation of [`HeaderPeeler`] against an array of bytes.
+pub fn try_extract_header<'a, P>(
+    bytes: &'a [u8],
```
SirCipher, line 79: It might be nice to make this `A: AsRef<&[u8]>` in case we need to use it with a `BytesMut` etc?

api/formats/swim_recon/src/parser/record/matcher/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+
+fn peel_message<'a, P>(
+    peeler: P,
+) -> impl FnMut(Span<'a>) -> IResult<Span<'a>, <P as HeaderPeeler<'a>>::Output>
+where
+    P: HeaderPeeler<'a>,
+{
+    move |input| {
+        map_res(
+            pair(peel_tag_attr(peeler.clone()), preceded(space0, rest)),
```
SirCipher, line 101: Not `multispace0`?

api/formats/swim_recon/src/parser/record/matcher/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+fn name(input: Span<'_>) -> IResult<Span<'_>, Cow<'_, str>> {
+    alt((map(identifier, Cow::Borrowed), string_literal))(input)
+}
+
+fn match_tag<'a, P>(peeler: P) -> impl FnMut(Span<'a>) -> IResult<Span<'a>, P>
+where
+    P: HeaderPeeler<'a>,
+{
+    move |input| {
+        map_res(preceded(char('@'), name), |tag| {
```
SirCipher, line 182: While it shouldn't happen, should this not have a `preceded(multispace0, ..)`?

api/formats/swim_recon/src/parser/record/matcher/tests.rs
```diff
@@ -0,0 +1,421 @@
...
+    "@attr(@inner)",
+    "@attr(@inner{})",
+    "{ @attr }",
+    "{ @attr, @attr(@inner) }",
+    "{{}}",
+    "{{{}}}",
+    "{{},{}}",
+    "@attr 4",
+    "@attr() 4",
+];
```
SirCipher, line 127: Is it worth adding some that test for trailing spaces?

api/formats/swim_recon/src/parser/record/matcher/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// The type of results produced after recognizing an attribute.
+    type Output: 'a;
+    /// Possible failure cases from attempting to interpret the attribute.
+    type Error: Clone + 'a;
+
+    /// Provide the name of the tag attribute.
+    ///
+    /// #Arguments
+    /// * `name` - The name of the tag. Not that this has a more restrictive lifetime as may
+    /// have been unescaped buy the parser.
```
DobromirM, line 48:
```suggestion
    /// * `name` - The name of the tag. Note that this has a more restrictive lifetime as may
    /// have been unescaped by the parser.
```

api/formats/swim_recon/src/parser/record/matcher/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    fn tag(self, name: &str) -> Result<Self, Self::Error>;
+
+    /// Feed a slot from the body of the attribute.
+    ///
+    /// #Arguments
+    /// * `name` - The name of the slot. Not that this has a more restrictive lifetime as may
+    /// have been unescaped buy the parser.
```
DobromirM, line 55:
```suggestion
    /// * `name` - The name of the slot. Note that this has a more restrictive lifetime as may
    /// have been unescaped by the parser.
```

api/formats/swim_recon/src/parser/record/matcher/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    /// Feed an empty item from the body of the attribute. (Note that the result of recognizing
+    /// "@attr" will be different from "@attr()").
+    fn feed_header_extant(self) -> Result<Self, Self::Error>;
+
+    /// Attempt to interpret the attribute.
+    ///
+    /// #Arguments
+    /// * `body` - The remainder of the input span. Note that this is entirely uninterpreted and
+    /// so many not contain valid Recon.
```
DobromirM, line 73:
```suggestion
    /// so may not contain valid Recon.
```

api/formats/swim_recon/src/parser/record/matcher/mod.rs
horned-sphere: New-lines aren't permitteed between an attribute and the remainder of a record (that tripped me up on my first recon parser).

DobromirM (approved):

api/formats/swim_recon/src/parser/record/matcher/mod.rs
horned-sphere: Annoyingly, it doesn't work. If you do that, you then need to strengthen `P: HeaderPeeler<'a>` to `P: for<'a> HeaderPeeler<'a>` and then you can't have a return type with `P::Output` as it's not as single type. (It's not that inconvenient though as it just means you have to call `as_ref()` to all it.

api/formats/swim_recon/src/parser/record/matcher/mod.rs
horned-sphere: I think I'd prefer it to be the callers responsibility to ensure that the record starts at the beginning of the input. Particularly as this is intended for use with warp frames rather than documents.

api/formats/swim_recon/src/parser/record/matcher/tests.rs
horned-sphere: This test case is just to verify that it can recognize value strings. The surroundings (including white space) are covered by other test cases so I don't think it needs to be added here (and would make the test check a lot messier).

SirCipher (approved):

PR #494, opened 2022-04-01
Map downlinks

DobromirM:


api/swim_form/src/structural/write/mod.rs
```diff
@@ -884,6 +884,40 @@ where
...
+        let len = self.len();
+        self.into_iter()
+            .try_fold(
+                writer
+                    .record(0)?
+                    .complete_header(RecordBodyKind::MapLike, len)?,
+                |record_writer, (key, value)| record_writer.write_slot_into(key, value),
+            )?
+            .done()
+    }
```
horned-sphere, line 918: Could this share code with the `HashMap` one? They're pretty much the same onece you get down to the iterator of entries.

swim_downlink/src/model/lifecycle/on_removed.rs
```diff
@@ -0,0 +1,169 @@
...
+use swim_api::handlers::{BlockingHandler, FnMutHandler, NoHandler, WithShared};
+
+/// Trait for event handlers to be called when a map entry is removed.
+pub trait OnRemoved<'a, K, V>: Send {
+    type OnRemovedFut: Future<Output = ()> + Send + 'a;
+
+    /// # Arguments
+    /// * `key` - The key of the map entry.
+    /// * `existing` - The existing value of the map entry, if it is defined.
+    fn on_removed(&'a mut self, key: &'a K, existing: Option<&'a V>) -> Self::OnRemovedFut;
```
horned-sphere, line 26: Once you  switch to `im::OrdMap`, you could provide a reference to the entire map here too.

swim_downlink/src/model/lifecycle/on_updated.rs
```diff
@@ -0,0 +1,201 @@
...
+
+/// Trait for event handlers to be called when a map entry is updated.
+pub trait OnUpdated<'a, K, V>: Send {
+    type OnUpdatedFut: Future<Output = ()> + Send + 'a;
+
+    /// # Arguments
+    /// * `key` - The key of the map entry.
+    /// * `existing` - The existing value of the map entry, if it is defined.
+    /// * `new_value` - The replacement value of the map entry.
+    fn on_updated(
```
horned-sphere, line 27: And again here, once you  switch to `im::OrdMap`, you could provide a reference to the entire map here too.

DobromirM: Unfortunately, I don't think there is a way to put the key in the map and also get a reference to it without cloning it

swim_downlink/src/model/lifecycle/on_removed.rs
DobromirM: `Take` and `Drop` also call `on_removed` but they have to iterate and remove entries from the map, making it impossible to borrow the map again for the lifecycle callback.

swim_downlink/src/model/mod.rs
```diff
@@ -60,6 +76,22 @@ pub type DefaultValueDownlinkModel<T> = ValueDownlinkModel<
...
+        V,
+        NoHandler,
+        NoHandler,
+        NoHandler,
+        NoHandler,
+        NoHandler,
+        NoHandler,
+        NoHandler,
+    >,
+>;
```
SirCipher, line 93: It would be tidier if `BasicMapDownlinkLifecycle` had default type parameters that delegate to `NoHandler`

swim_downlink/src/task/map.rs
```diff
@@ -0,0 +1,272 @@
...
+    let read = async move {
+        let result = read_task(config, input, lifecycle).await;
+        let _ = stop_tx.trigger();
+        result
+    }
+    .instrument(info_span!("Downlink read task.", %path));
+    let framed_writer = FramedWrite::new(output, DownlinkOperationEncoder);
+    let write = write_task(framed_writer, event_stream, stop_rx)
+        .instrument(info_span!("Downlink write task.", %path));
+    let (read_result, _) = join(read, write).await;
```
SirCipher, line 76: If the write task encounters an error, it will stop but the read task will continue. Should they not both stop together?

swim_downlink/src/task/map.rs
```diff
@@ -0,0 +1,272 @@
...
+            }
+
+            map = map_iter.map(|(_, (key, value))| (key, value)).collect();
+        }
+        MapMessage::Drop(n) => {
+            let mut map_iter = map.into_iter().rev().enumerate();
+
+            for (idx, (key, prev_value)) in &mut map_iter {
+                if with_callbacks {
+                    lifecycle.on_removed(&key, Some(&prev_value)).await;
```
SirCipher, line 239: It would be more efficient to add a handler for `on_dropped(OrdMap)` so this could be done in a single call

swim_downlink/src/task/map.rs
```diff
@@ -0,0 +1,272 @@
...
+
+                if n - 1 <= idx as u64 {
+                    break;
+                }
+            }
+
+            map = map_iter.map(|(_, (key, value))| (key, value)).collect();
+        }
+        MapMessage::Drop(n) => {
+            let mut map_iter = map.into_iter().rev().enumerate();
```
SirCipher, line 235: This can be just `iter` and then the map doesn't need to be reassigned. Same for `Take`

swim_downlink/src/task/map.rs
SirCipher: Also, as this doesn't actually drop/take any elements, should a log message be printed to say that it's not implemented?

swim_downlink/src/task/mod.rs
```diff
@@ -85,8 +88,34 @@ where
...
             .boxed()
     }
 }
+
+impl<K, V, LC> Downlink for DownlinkTask<MapDownlinkModel<K, V, LC>>
+where
+    K: Form + Ord + Eq + Clone + Send + Sync + 'static,
+    V: Form + Clone + Send + Sync + 'static,
+    <K as RecognizerReadable>::Rec: Send,
+    <V as RecognizerReadable>::Rec: Send,
```
SirCipher, line 102: This would be tidier if `K/V: RecognizerReadable` and then `K/V::Rec: Send`

swim_downlink/src/model/lifecycle/mod.rs
```diff
@@ -1084,3 +1134,1181 @@ where
...
+    on_synced: FSync,
+    on_event: FEv,
+    on_updated: FUpd,
+    on_removed: FRmv,
+    on_cleared: FClr,
+    on_unlinked: FUnlink,
+}
+
+/// Create a default lifecycle for a map downlink (where all of the event handlers do nothing).
+pub fn for_map_downlink<K, V>() -> BasicMapDownlinkLifecycle<
```
SirCipher, line 1152: If you applied my other comment about adding the default type parameters for `BasicMapDownlinkLifecycle` this could be done as a `Default` implementation for only the `K,V` bounds

swim_downlink/src/model/mod.rs
horned-sphere: This is my fault but fair :)

PR #496, opened 2022-04-08
Changed to the encoder and decoder to handle map messages.

DobromirM:


api/swim_api/src/protocol/downlink/mod.rs
```diff
@@ -57,7 +67,25 @@ use super::{
...
+#[derive(Debug, Clone, Copy)]
+pub struct SimpleMessageEncoder;
+
+impl<T: AsRef<[u8]>> Encoder<T> for SimpleMessageEncoder {
+    type Error = std::io::Error;
+
+    fn encode(&mut self, item: T, dst: &mut BytesMut) -> Result<(), Self::Error> {
+        let item_bytes = item.as_ref();
+        let item_len = item_bytes.len();
+        dst.reserve(TAG_SIZE + item_len);
```
horned-sphere, line 79: You're not writing a tag here so `TAG_SIZE` isn't needed.

horned-sphere (commented):
I think the size in the value frames is better for robustness.

PR #500, opened 2022-07-14
Derive macro for AgentLaneModel.

horned-sphere:


server/swim_agent/tests/deriveagentlanemodel.rs
```diff
@@ -0,0 +1,162 @@
...
+    struct MultipleLanes {
+        first: ValueLane<i32>,
+        second: MapLane<i32, i32>,
+        third: ValueLane<i32>,
+        fourth: MapLane<i32, i32>,
+        fifth: CommandLane<i32>,
+    }
+
+    check_agent::<MultipleLanes>(vec!["first", "third", "fifth"], vec!["second", "fourth"]);
+}
```
SirCipher, line 162: I would also add in a test for a lane that doesn't depend on an import; `lane: swim_agent::lanes::CommandLane`

SirCipher: As well as deriving two agent lane models in the same scope

server/swim_agent_derive/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+    let input = parse_macro_input!(input as DeriveInput);
+    lane_model_derive::validate_input(&input)
+        .map(DeriveAgentLaneModel::new)
+        .map(ToTokens::into_token_stream)
+        .into_result()
+        .unwrap_or_else(errs_to_compile_errors)
+        .into()
+}
+
+fn errs_to_compile_errors(errors: Errors<syn::Error>) -> proc_macro2::TokenStream {
```
SirCipher, line 35: There is a function like this in the macro utilities crate that accepts the vectorized errors

SirCipher (approved):

PR #502, opened 2022-07-22
Attribute macro to derive agent lifecycles.

horned-sphere:
For examples of using the macro, see server/swim_agent/tests/agentlifecycle.rs.


server/swim_agent_derive/src/lib.rs
```diff
@@ -55,3 +57,25 @@ pub fn projections(attr: TokenStream, item: TokenStream) -> TokenStream {
         .unwrap_or_else(|errs| to_compile_errors(errs.into_vec()))
         .into()
 }
+
+#[proc_macro_attribute]
+pub fn lifecycle(attr: TokenStream, item: TokenStream) -> TokenStream {
```
SirCipher, line 62: It would be good to add some documentation to this

horned-sphere: Do you mean internal documentation to explain how the macro works or documentation in the crate to explain how to use it? (I know the answer is both, but I think the second one is quite a big job and should probably be its own PR).

SirCipher: More the latter. Maybe it's best to make a ticket for it and circle back to it later

horned-sphere: I'll add the ticket and work on it fairly soon. I think it makes sense to do some consolidation now we're back to parity(ish) with the prototype code.

SirCipher (approved):

PR #525, opened 2022-10-20
Adds support for agent lane persistence.

horned-sphere:
Built on top of the storage enging in the prototype.


api/swim_api/src/store.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    /// Attempt to copy the single value associated with an ID into the provided buffer. The value associated
+    /// with the id should be copied into the buffer (leaving any existing content intact) and the number
+    /// of bytes copied should be returned. If no data is associated with the key the result should be valid
+    /// and undefined.
+    fn get_value(
+        &self,
+        id: Self::LaneId,
+        buffer: &mut BytesMut,
+    ) -> Result<Option<usize>, StoreError>;
```
SirCipher, line 60: What do you think about using a `ReadBuf` here instead? It would be more inline with this sort of flow.

api/swim_model/src/address.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub host: Option<T>,
+    pub node: T,
+    pub lane: T,
+}
+
+impl<T: Display + Debug> Display for Address<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Address(host = {:?}, ode = {}, lane = {})",
```
SirCipher, line 36: *node

SirCipher (approved):

api/swim_api/src/store.rs
horned-sphere: I wanted to keep it simple to implement (so it can't fail because the buffer doesn't have enough space). Also it's only called internally where it will always be used with a `BytesMut` rather than any other type of buffer.

PR #527, opened 2022-11-14
Adds support for agent and lane introspection.

horned-sphere:


server/swim_introspection/src/meta_agent/node/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+) -> AgentInitResult {
+    let pattern = node_pattern();
+    let params = match pattern.unapply_route_uri(&route) {
+        Ok(params) => params,
+        Err(e) => return Err(AgentInitError::UserCodeError(Box::new(e))),
+    };
+    let node_uri = &params[NODE_PARAM];
+    let handle = match resolver.resolve_agent(Text::from(node_uri)).await {
+        Ok(handle) => handle,
+        Err(e) => return Err(AgentInitError::UserCodeError(Box::new(e))),
```
SirCipher, line 107: These aren't user-created errors, are they? It be better to add another variant so that we can discriminate against resolution errors?

server/swim_introspection/src/meta_agent/node/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+    let pulse_io = context
+        .add_lane(PULSE_LANE, LaneKind::Supply, lane_config)
+        .await?;
+    let lanes_io = context
+        .add_lane(LANES_LANE, LaneKind::DemandMap, lane_config)
+        .await?;
+    Ok(run_task(pulse_interval, handle, pulse_io, lanes_io).boxed())
+}
+
+type Io = (ByteWriter, ByteReader);
```
SirCipher, line 120: I really should've added a ByteChannel type to the crate itself as we write this type a lot.

server/swim_introspection/src/task/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+/// * `stopping` - Signal that the server is stopping.
+/// * `channel_size` - Size of the channel use to register new lanes.
+pub fn init_introspection(
+    stopping: trigger::Receiver,
+    channel_size: NonZeroUsize,
+) -> (
+    IntrospectionResolver,
+    impl Future<Output = ()> + Send + 'static,
+) {
+    let (msg_tx, msg_rx) = mpsc::unbounded_channel();
```
SirCipher, line 101: Why unbounded?

server/swim_introspection/src/meta_agent/node/mod.rs
horned-sphere: Probably a good idea to add this. It would be messy getting rid of all the existing aliases though.

server/swim_introspection/src/task/mod.rs
horned-sphere: There is a both a technical reason and an _a posteriori_ excuse!

Technically, the `Agents` struct is not `Sync` which means the `resolve_agent` method can't be made `async` without making the top level future `Send` which breaks everything. In principle, things could be restrucutured so that this can work but I don't think it would solve the fundamental problem.

The excuse part is that I don't think we can afford this task to lock up an wait. The alternative would be to shove the `resolve_agent` operation off into a `FuturesUnordered` on the side but that just moves the unbounded queue from the channel to the `FuturesUnordered`. Fundamentally, starting up agents _is_ unbounded as clients can request the arbitrarily rapidly. We might want to consider what we should do in this situation (an agent DOS attack essentially) but I think it would be a task in itself.

SirCipher (approved):

PR #530, opened 2022-11-22
Co-op budget for byte channels.

horned-sphere:
Ideally, we would integrate the byte channels with the tokio co-op mechanism, however tokio makes this very difficult. It only provides an unstable async function that allows directly consuming the co-op budget.

I experimented with using `tokio::sync::PollSemaphore` rather than a parking_lot `Mutex` in the channels but this made them significantly slower.

This PR introduces a separate co-op system for the byte channels. (Interestingly, the benchmarks now don't show much benefit to using our channels as compared to the tokio equivalent so I think a lot of our benefit may have come from not periodically yeielding, depressingly!).


SirCipher:
I feel that this should be disabled by default and for us to then have a global 'unstable' (or similar) flag that we can use to enable features like this as well as a 'coop' flag. 

Alongside this, would it not be more beneficial to spawn the tasks (in as small a scope as possible) that use this budgeting system as unconstrained to prevent the yielding occurring too frequently - as the code in this PR will cause the task to yield when its budget is exhausted and then when the Tokio task's operation budget is exhausted that will also cause a yield? 

It would also be interesting to see how this approach fairs to using the unstable 'consume_budget' function

horned-sphere:
> I feel that this should be disabled by default and for us to then have a global 'unstable' (or similar) flag that we can use to enable features like this as well as a 'coop' flag.

The problem is, I don't think it would be wise to have this off. If you have a lot of agents running at the same time and one of them is getting constant traffic it would starve the others.

> 
> Alongside this, would it not be more beneficial to spawn the tasks (in as small a scope as possible) that use this budgeting system as unconstrained to prevent the yielding occurring too frequently - 

As it is, the scope is on a per-agent basis (which is essentially the conept of a `task` in a swim application). I'm not sure how it could be made much smaller.

>as the code in this PR will cause the task to yield when its budget is exhausted and then when the Tokio task's operation budget is exhausted that will also cause a yield?
> 

Yes, unfortunately that's not really avoidable. It wouldn't work to unconstrain the tasks as they could be busy from traffic over tokio channels as well as over our byte channels and so that would give anotehr route to starvation.

> It would also be interesting to see how this approach fairs to using the unstable 'consume_budget' function

I did consider that (as a less bad option), however it's not easy to do. The channels are used via `AsyncRead` an `AsyncWrite`. You can't call into `consume_budget` from these as it's an `async` function. If we wanted to go down this route, we'd need to stop using `FramedRead` and `FramedWrite` and write our own implementations that call `consume_budget` each time they do a successful IO operation. An alternative is to wait for TAIT (which is going to be stablized soon) which would allow us to use the `consume_budget` function in the `AsyncRead` and `AsyncWrite` impls.

PR #541, opened 2023-01-16
Store implementation that holds agent state in memory.

horned-sphere:
This store allows agent state to persist across resets (but not when the process stops).

All state is held in memory.


runtime/swim_persistence/src/agent/in_memory/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+        let mut guard = plane.lock();
+        let map = &mut guard.nodes;
+        if let Some(NodeEntry::InUse(Some(tx))) = map.remove(uri) {
+            if let Err(state) = tx.send(state) {
+                map.insert(std::mem::take(uri), NodeEntry::Idle(state));
+            } else {
+                map.insert(std::mem::take(uri), NodeEntry::InUse(None));
+            }
+        } else {
+            map.insert(std::mem::take(uri), NodeEntry::Idle(state));
```
SirCipher, line 297: It shouldn't be possible for this branch to be reached, should it?

runtime/swim_persistence/src/agent/in_memory/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+        Ok(())
+    }
+
+    fn clear_map(&mut self, id: Self::LaneId) -> Result<(), StoreError> {
+        let InMemoryNodePersistence {
+            state: NodeState { maps, values, .. },
+            ..
+        } = self;
+        if maps.remove(&id).is_none() && values.contains_key(&id) {
+            Err(StoreError::InvalidOperation)
```
SirCipher, line 263: Should operations like these not be considered bugs more than anything else?

runtime/swim_persistence/src/agent/in_memory/tests.rs
```diff
@@ -0,0 +1,366 @@
+// Copyright 2015-2021 Swim Inc.
```
SirCipher, line 1: Would be worth modifying the persistence integration tests to run on both the rocks store and this in-memory store to check that they both yield the same results?

swim_store/rocks_store/src/lib.rs
```diff
@@ -42,6 +43,42 @@ impl RocksEngine {
...
+    }
+
+    pub fn prefix_iterator_opt<'a: 'b, 'b>(
+        &'a self,
+        space: &'b <Self as KeyspaceResolver>::ResolvedKeyspace,
+        prefix: &'b [u8],
+    ) -> Result<RocksPrefixIterator<'b>, StoreError> {
+        let it = self.delegate.prefix_iterator_cf(space, prefix);
+        Ok(it.into())
+    }
```
SirCipher, line 81: Two of these functions are missing the read options

runtime/swim_persistence/src/agent/in_memory/mod.rs
horned-sphere: The way it is writte now, no it can't ocurr. However, the only alternative would be to panic which is adding a potential nasty surprise if the implementation changes later (panicking destructors are not much fun).

runtime/swim_persistence/src/agent/in_memory/mod.rs
horned-sphere: Ideally, there would be different types of ID for different store types so they can't be mixed up. I started doing it but it degenerated into a mess so I decided to leave it on the side for now.

swim_store/rocks_store/src/lib.rs
SirCipher: I'm pretty sure that the `_opt` functions used to also have an argument of `opts: ReadOptions` that was passed to the delegate calls. Such as `raw_iterator_cf_opt`. As at the moment, two of the functions just delegate to each other

horned-sphere: True, and they're not used. I think I was in the process of refactoring them out and never actually deleted them. I've removed them now.

runtime/swim_persistence/src/agent/in_memory/tests.rs
horned-sphere: Done. They _mostly_ did.

SirCipher (approved):

PR #542, opened 2023-01-18
Adds a value store type.

horned-sphere:
Support for stores in the the runtime and derivations macros will be added later.


server/swim_agent/src/stores/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use bytes::BytesMut;
+
+use crate::{agent_model::WriteResult, AgentItem};
+
+pub mod value;
+
+pub trait Store: AgentItem {
```
SirCipher, line 21: Would it be better if this was named something else so it doesn't collide with the other store trait? Maybe LaneStore?

server/swim_agent/src/stores/value/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+
+/// Adding a [`ValueStore`] to an agent provides additional state that is not exposed as a lane.
+/// If persistence is enabled (and the store is not marked as transient) the state of the store
+/// will be persisted in the same was as the state of a lane.
+#[derive(Debug)]
+pub struct ValueStore<T> {
+    id: u64,
+    content: RefCell<T>,
+    previous: RefCell<Option<T>>,
+    dirty: Cell<bool>,
```
SirCipher, line 43: Is it worth having three cells here? Could the inners not be within one RefCell?

server/swim_agent/src/lib.rs
```diff
@@ -34,6 +31,10 @@ pub mod model {
     pub use swim_model::Text;
 }
 
+pub trait AgentItem {
```
SirCipher, line 34: Could you add some documentation to this?

api/swim_api/src/protocol/agent/store/tests.rs
```diff
@@ -0,0 +1,159 @@
...
+
+#[test]
+fn encode_command_store_request() {
+    let mut encoder = StoreInitMessageEncoder::value();
+    let mut buffer = BytesMut::new();
+    let content = b"body";
+    let message = StoreInitMessage::Command(content);
+    assert!(encoder.encode(message, &mut buffer).is_ok());
+
+    assert_eq!(buffer.remaining(), 9 + content.len());
```
SirCipher, line 39: Could you move these '9' magic numbers to a const?

server/swim_agent/src/agent_model/init/mod.rs
```diff
@@ -175,40 +170,60 @@ impl<Agent, K, V> MapLaneInitializer<Agent, K, V> {
...
+) -> Result<InitFn<Agent>, FrameIoError>
+where
+    Agent: 'static,
+    T: RecognizerReadable + Send + 'static,
+    F: FnOnce(&Agent, T) + Send + 'static,
+{
+    let mut body = None;
+    while let Some(result) = stream.next().await {
+        body = Some(result?);
+    }
```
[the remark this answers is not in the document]
horned-sphere: I've added a utlity method for it. I've only replaced this one usage though as it's not a pattern that's easy to search for. It can be replaced if we notice more in future.

server/swim_agent/src/stores/mod.rs
horned-sphere: I've changed Store -> StoreItem and Lane to LaneItem.

horned-sphere: And it's going to cause lots of merge conflicts later 😭.

server/swim_agent/src/stores/value/mod.rs
horned-sphere: I've combined the two RefCells. It's actually better for the dirty cell to be separate (it has 0 overhead and doesn't require a borrow when it gets read).

server/swim_agent/src/stores/mod.rs
SirCipher: 🥲

server/swim_agent/src/stores/value/mod.rs
SirCipher: That's fair. I just thought for the RefCells it would save a bit of space

SirCipher (approved):

PR #544, opened 2023-01-23
Client runtime

SirCipher:


api/swim_form/src/tests/a.rs
```diff
@@ -0,0 +1 @@
+
```
horned-sphere, line 1: What's this?

client/runtime/src/downlink/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+use swim_utilities::trigger::promise;
+
+pub type DownlinkOperationResult<T> = Result<T, DownlinkRuntimeError>;
+
+/// A view over a stateful downlink.
+#[derive(Debug, Clone)]
+pub struct StatefulDownlinkView<T> {
+    // This is intentionally its own field as to access it in the watch channel's receiver requires
+    // accessing its internal mutex.
+    pub current: Arc<T>,
```
horned-sphere, line 30: It shouldn't be necessary for the downlink state to be in an `Arc`. If you makes sure the downlink just runs within a single task and are very careful with how you specify `Send` and `Sync` bounds you can generally just have `T` as the state (or posisbly somtimes you might need to drop to `RefCell`).

I think, in theory, you should be able to use the stuff that I already wrote for this.

client/runtime/src/runtime/mod.rs
```diff
@@ -0,0 +1,715 @@
...
+    (
+        RawHandle {
+            dispatch: requests_tx,
+            _task,
+        },
+        remote_stop_tx,
+    )
+}
+
+const TRANSPORT_BUFFER_SIZE: NonZeroUsize = non_zero_usize!(32);
```
horned-sphere, line 216: Why is this fixed?

client/runtime/src/runtime/models.rs
```diff
@@ -0,0 +1,225 @@
...
+use swim_utilities::io::byte_channel::{ByteReader, ByteWriter};
+use swim_utilities::trigger;
+use tokio::sync::mpsc;
+use uuid::Uuid;
+
+#[derive(Debug, Hash, Clone, Eq, PartialEq)]
+pub struct Key {
+    pub address: RelativeAddress<Text>,
+    pub kind: DownlinkKind,
+    pub config_hash: u64,
```
horned-sphere, line 34: What is this used for?

client/swim_client/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+        SwimClient {
+            stop_tx,
+            handle: ClientHandle {
+                inner: Arc::new(handle),
+            },
+        }
+    }
+
+    /// Returns a reference to a cloneable handle which may be used to open downlinks.
+    pub fn handle(&self) -> &ClientHandle {
```
horned-sphere, line 71: It might be nicer to just make this return by value.

client/swim_client/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+        &self.handle
+    }
+
+    /// Triggers the runtime to shutdown.
+    pub fn shutdown(self) {
```
horned-sphere, line 76: Should this not be async to allow for a clean shutdown?

client/runtime/src/runtime/models.rs
SirCipher: This is for per-downlink configurations. Allowing for multiple downlinks to the same address, kind, and configuration to share the same runtime.

client/swim_client/src/lib.rs
SirCipher: This fires the trigger inside - which isn't an async call.

horned-sphere: No, but the client runtime will still have (async) work to do and you'd generally want to wait for it to complete before letting the process end.

client/runtime/src/runtime/models.rs
horned-sphere: I can kind of see the argument for that in a server (where there might be independent agents wanting a downlink to the same place). It seems a bit over the top for a clien though. Also, you could (implausibly, admitedly) have collisons. If you want to do this, you could maybe use time based UUIDs.

client/runtime/src/models.rs
```diff
@@ -27,35 +25,7 @@ use swim_utilities::trigger;
...
-    pub fn parts(&self) -> (&RelativeAddress<Text>, DownlinkKind, u64) {
-        let Key {
-            address,
-            kind,
-            config_hash,
-        } = self;
-        (address, *kind, *config_hash)
-    }
-}
+pub type Key = (RelativeAddress<Text>, DownlinkKind);
```
horned-sphere, line 28: I think having the struct made sense, even without the hash (no need to add it back if you prefer this though).

client/swim_client/src/lib.rs
```diff
@@ -57,24 +63,31 @@ impl SwimClient {
             subprotocols: Default::default(),
         };
         let networking = TokioPlainTextNetworking::new(Arc::new(Resolver::new().await));
-        let (handle, stop_tx) =
```
horned-sphere: I just noticed that the line above misses out TLS support.

swim_utilities/swim_future/src/combinators/race/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+    L: Unpin,
+    R: Unpin,
+{
+}
+
+/// Waits for either one of two differently-typed futures to complete and then discards the other
+/// future. If both futures are ready then the left future's output is returned.
+///
+/// This function differs to Future's Select functionality where it discards the remaining future.
+pub fn race<L, R>(left: L, right: R) -> Race2<L, R>
```
horned-sphere, line 28: You might want to add something about cancel safety here. Passing IO related futures to this would be a bad idea.

client/swim_client/src/lib.rs
SirCipher: I've added trust-dns, deflate and tls support.

horned-sphere (approved):

PR #556, opened 2023-04-04
Unifies value and map item specifications in the agent model

SirCipher:


example_apps/console/src/ui/mod.rs
```diff
@@ -157,6 +157,8 @@ pub fn create_ui(siv: &mut Cursive, mut controller: Controller, max_lines: usize
                             .on_submit_mut(move |s, text| {
                                 on_command(s, &mut controller, &history_appender, text)
                             })
+                            .scrollable()
```
horned-sphere, line 160: This element is the box where the commands get typed. It shouldn't be scrollable.

runtime/swim_runtime/src/agent/task/mod.rs
```diff
@@ -936,6 +935,37 @@ enum WriteTaskEvent<I> {
     Stop,
 }
 
+impl<I> Debug for WriteTaskEvent<I> {
```
horned-sphere, line 938: What's wrong with the derived one?

server/swim_agent/src/agent_model/tests/fake_agent.rs
```diff
@@ -146,6 +142,32 @@ impl AgentSpec for TestAgent {
...
+    where
+        Self: 'static,
+    {
+        None
+    }
+
+    fn init_map_like_item(
+        &self,
+        _item: &str,
+    ) -> Option<
```
horned-sphere, line 158: This type is a bit of a mouthful. Have you run clippy? I suspec it would complain.

server/swim_agent/src/lanes/join_value/downlink/tests.rs
```diff
@@ -424,6 +424,7 @@ fn run_on_failed_abandon() {
 }
 
 #[test]
+#[ignore]
```
horned-sphere, line 427: Make a ticket for this so I don't forget it.

runtime/swim_runtime/src/agent/task/mod.rs
SirCipher: Where it's printed, it requires AgentPersistence::StoreId: Debug -which it isn't constrained by but could be added as they're all either units or u64s.

server/swim_agent/src/agent_model/tests/fake_agent.rs
SirCipher: It was rustfmt that shifted this up from lower in the file. Surprisingly, clippy doesn't complain about it

PR #557, opened 2023-04-04
Defers drop of agent context in meta node agent

SirCipher:


example_apps/basic/src/main.rs
```diff
@@ -38,6 +38,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
         .update_config(|config| {
             config.agent_runtime.inactive_timeout = Duration::from_secs(5 * 60);
         })
+        .enable_introspection()
```
horned-sphere, line 41: You should make a new example app for demonstrating introspection. That can go its own PR though.

server/swim_introspection/src/meta_agent/node/mod.rs
```diff
@@ -154,7 +155,10 @@ async fn run_task(
                 error,
             })
         }
-    }
+    };
+    // deferred drop so the agent doesn't terminate early.
+    let _context = context;
```
horned-sphere, line 160: This should go at the top rather than the end. It looks a bit weird assigning it immediately before it gets dropped.

horned-sphere (commented):
I think, ultimeately, it should be changed so that you don't have to keep hold of the context if you don't need it (the runtime could keep hold of a handle maybe). This works for now though.

SirCipher:
@horned-sphere, that's a valid point. The context trait doesn't actually require any mutable reference to itself so clone could be implemented for it and then the runtime could keep a reference to it - happy to make that change if you'd prefer it?

horned-sphere:
> @horned-sphere, that's a valid point. The context trait doesn't actually require any mutable reference to itself so clone could be implemented for it and then the runtime could keep a reference to it - happy to make that change if you'd prefer it?

The actual implementation of the contect just has an mpsc sender it. The runtime could just keep hold of a copy of the sender rather than making the context cloneable. It _might_ have unanticipated effects though (you might need to make sure the sender is deliberately dropped at some point to make sure the shutdown works as it should). You could have a go but I think this solutio is fine. Maybe make a ticket for changing it later?

SirCipher:
I'll make a ticket for doing it later and make the changes that you've requested here. Regarding your comment, could the runtime not keep a strong reference to the sender and the context have a weak one that's upgraded each time it's required?

horned-sphere:
> I'll make a ticket for doing it later and make the changes that you've requested here. Regarding your comment, could the runtime not keep a strong reference to the sender and the context have a weak one that's upgraded each time it's required?

I don't think that should be necessary. If things are working properly, the runtime task will always stop last.

horned-sphere (approved):

PR #561, opened 2023-04-14
Adds event downlink example.

horned-sphere:
Fixes some bugs around downlink handling.
API quaility improvements to make working with downkinks and state less verbose.
Adds more logging to the downlink connector task to help with debugging.


docs/lifecycle.md
```diff
@@ -248,14 +249,16 @@ Interior mutability in lifecycles
 
 The reference to `self` for the agent lifecycle is always passed as `&self` and a function taking `&mut self` cannot be used for an event handler. This is as an event handler can trigger other handlers on the same lifecycle (even itself) which results in multiple borrows of the data in the lifecycle.
 
+Often, it will not be possible for a stateful lifecycle to be `Clone` (as the state may not be cloneable). In such a case, the `no_clone` option to the lifecycle macro. When registering the agent i the server, it will be necessary to provide a factory rather than a template instance.
```
SirCipher, line 252:
```suggestion
Often, it will not be possible for a stateful lifecycle to be `Clone` (as the state may not be cloneable). In such a case, the `no_clone` option to the lifecycle macro. When registering the agent in the server, it will be necessary to provide a factory rather than a template instance.
```

example_apps/example_util/src/lib.rs
```diff
@@ -63,3 +65,85 @@ impl<'a, K: Display, V: Display> Display for FormatMap<'a, K, V> {
...
+            filter
+        } else {
+            EnvFilter::new("")
+                .add_directive("swim_server_app=trace".parse()?)
+                .add_directive("swim_runtime=trace".parse()?)
+                .add_directive("swim_agent=trace".parse()?)
+                .add_directive("swim_messages=trace".parse()?)
+                .add_directive("swim_remote=trace".parse()?)
+                .add_directive("mio=warn".parse()?)
+                .add_directive("tokio=warn".parse()?)
```
SirCipher, line 143: Should these be in an example? I can't see why mio would need to be

SirCipher (approved):

example_apps/example_util/src/lib.rs
horned-sphere: I've simplified it a bit.

SirCipher (approved):

PR #567, opened 2023-05-16
Adds support for sending ad hoc commands from agents.

horned-sphere:


api/swim_model/src/bytes_str.rs
```diff
@@ -82,3 +84,25 @@ impl PartialEq<String> for BytesStr {
         self.as_ref() == other
     }
 }
+
+pub trait TryFromUtf8Bytes: Sized {
+    fn try_from_utf8_bytes(bytes: Bytes) -> Result<Self, std::str::Utf8Error>;
```
SirCipher, line 89: This would be more ergonomic if it accepted `&[u8]` as it would save on freezing the buffer before the call and then calling `as_ref` on it

horned-sphere: The reason it takes `Bytes` is to avoid a copy for the case of `BytesStr` (which can just consume the `Bytes` object).

SirCipher (approved):

PR #569, opened 2023-05-19
Reworks msgpack implementation to support delegate body operations

SirCipher:


api/formats/swim_msgpack/src/writer/mod.rs
```diff
@@ -12,94 +12,415 @@
...
+    /// Allocates a buffer with a capacity of 'len' for use by a closure that writes a primitive
+    /// value.
+    ///
+    /// Returns a result containing the written data or a message pack write error.
+    fn write<F, E>(&self, len: usize, f: F) -> Result<Bytes, MsgPackWriteError>
+    where
+        F: FnOnce(&mut ByteBuf) -> Result<(), E>,
+        MsgPackWriteError: From<E>,
+    {
+        let mut buf = ByteBuf::with_capacity(len);
```
horned-sphere, line 73: Allocating a new buffer for every value written seems fairly inefficent. Why did you make this change?

horned-sphere (approved):

PR #570, opened 2023-06-19
Adds support for keeping downlinks linked after failure.

horned-sphere:


api/swim_api/src/agent/mod.rs
```diff
@@ -142,12 +143,14 @@ pub trait AgentContext: Sync {
 #[derive(Debug, Clone, Copy)]
 pub struct AgentConfig {
     pub default_lane_config: Option<LaneConfig>,
+    pub keep_linked_retry: RetryStrategy,
 }
 
 impl AgentConfig {
     //TODO: Remove this once const impls are stable.
     pub const DEFAULT: AgentConfig = AgentConfig {
```
SirCipher, line 151: `RetryStrategy::none` is a const fn, so this should be possible without `const impl`?

SirCipher: I think that there is another comment for this elsewhere too

server/swim_agent/src/agent_model/downlink/hosted/event/mod.rs
```diff
@@ -38,98 +49,177 @@ use super::{DlState, DlStateObserver, DlStateTracker};
...
-                        biased;
-                        triggered_result = stop_signal => {
-                            *stop_rx = None;
-                            if triggered_result.is_ok() {
-                                *receiver = None;
-                                Some(Ok(DownlinkNotification::Unlinked))
+        if !*write_terminated {
+            *write_terminated = true;
+            return Some(Ok(DownlinkChannelEvent::WriteStreamTerminated));
+        }
```
SirCipher, line 144: It isn't immediately clear as to what's going on here, could you add a comment?

api/swim_api/src/agent/mod.rs
horned-sphere: It could be done with a `const` factory like you say. That doesn't seem really preferable to just having the `const` instance though.

SirCipher (approved):

PR #574, opened 2023-07-18
Adds hyper web server into the swim server.

horned-sphere:


server/swim_server_app/Cargo.toml
```diff
@@ -31,6 +32,8 @@ url = { workspace = true }
 swim_persistence = { path = "../../runtime/swim_persistence" }
 parking_lot = { workspace = true }
 rocksdb = { workspace = true, optional = true }
+hyper = { workspace = true, features = ["full", "tcp", "http1", "backports"] }
```
SirCipher, line 35: Do we need 'full' here?

server/swim_server_app/src/server/runtime/mod.rs
```diff
@@ -335,7 +345,10 @@ where
         let mut cmd_connection_tasks = FuturesUnordered::new();
         let mut client_tasks = FuturesUnordered::new();
 
-        let mut accept_stream = listener.into_stream().take_until(stop_signal);
+        let mut web_server = websockets
+            .wrap_listener(listener, ext_provider.clone())
+            .take_until(stop_signal);
+        //let mut accept_stream = listener.into_stream().take_until(stop_signal);
```
SirCipher, line 351: Rogue comment

server/swim_server_app/src/server/runtime/tests/connections.rs
```diff
@@ -56,14 +59,18 @@ pub struct TestWs {
...
-        socket: DuplexStream,
+        socket: Sock,
+        _provider: &'a Provider,
         _addr: String,
-    ) -> WsOpenFuture<DuplexStream, Self::Ext, RatchetError> {
+    ) -> WsOpenFuture<'a, Sock, Provider::Extension, RatchetError>
+    where
+        Sock: io::AsyncRead + io::AsyncWrite + Send + Unpin + 'static,
+        Provider: ratchet::ExtensionProvider + Send + Sync + 'static,
+        Provider::Extension: Send + Sync + 'static,
```
SirCipher, line 72: These bounds are in quite a few places. Is it possible to tie them to a trait to save this?

server/swim_server_app/src/server/http/mod.rs
```diff
@@ -0,0 +1,650 @@
...
+                .serve_connection(sock, svc)
+                .with_upgrades()
+        },
+    );
+
+    futures::stream::unfold(state, |mut state| async move {
+        state.next().await.map(move |result| (result, state))
+    })
+}
+
```
SirCipher, line 93: Could you add some comments to the rest of the file? Maybe a high-level explanation of the implementation?

server/swim_server_app/src/server/runtime/tests/connections.rs
horned-sphere: Yes, but it wasn't particularly straightforward...

SirCipher: Leave it if it's a hassle :)

horned-sphere: I had to roll that back as it causes a complete mess with the compiler failing to track the bounds through on the Provider associated type. I don't think it's really feasible without making changes to ratchet (or waiting for the rust compiler to get smarter). The best I could manage was:

```rust
pub trait ShareableExtensionProvider: Send + Sync + Unpin + 'static {
    type Extension: Extension + Send + Sync + Unpin + 'static;
    type Provider: ExtensionProvider<Extension = Self::Extension> + Send + Sync + Unpin + 'static;
    
    fn as_provider(&self) -> &Self::Provider;
    
    fn into_provider(self) -> Self::Provider;
}

impl<P> ShareableExtensionProvider for P
where
    P: ExtensionProvider + Send + Sync + Unpin + 'static,
    P::Extension:  Extension + Send + Sync + Unpin + 'static,
{
    type Provider = P;
    type Extension = P::Extension;
    
    fn as_provider(&self) -> &Self::Provider {
        self
    }
    
    fn into_provider(self) -> Self::Provider {
        self
    }
}
```

It works for some things but seems to really confuse the compiler with async blocks.

server/swim_server_app/src/server/http/mod.rs
horned-sphere: I've added more documentation to this. Whilst reading through it again, I've realised it doesn't work exactly as I inteded. However, it gets modified quite a bit in the next PR so if I change it now it will cause merge chaos. I'll fix it up on the next one instead.

SirCipher (approved):

PR #579, opened 2023-09-18
Adds the ability to rename lanes of agents in the derive macro.

horned-sphere:
This allows agent item names to be renamed explicitly or by convention. A convention may be applied at the agent or item level (directives on items override those defined for the agent).

These improvements are also backported to the Form derivation macro.


SirCipher:
As far as I can tell, this allows for an input of:
```rust
struct S {
        #[form(convention = "camel")]
        second_field: i32,
        secondField: String,
    }
```

To produce duplicate fields in the resulting record. It may be worth adding some checks for field name uniqueness in the derivation after the renaming has been applied.

Also, I think that it would be best to default the derived field names to be camelcase unless overridden. This will allow for the Java and Rust runtimes to communicate cleanly without the need for this renaming convention to be applied.

SirCipher:
@horned-sphere, what do you think about making camel case the default format applied to the fields?

SirCipher (approved):

PR #581, opened 2023-09-29
Supply lanes

### LINKED ISSUE #507: Add support for supply lanes.
#506 is a prerequisite for this issue.

Add supply lanes to `swim_agent::lanes` (see existing value and map lanes in this module for comparison).

Supply lanes should behave in the same way to their Java equivalent.

A demand lane has an associated type `T` and has no associated event handlers.

There should be an associated `EventHandler` which pushes a value of type `T` into the lane. When this is executed, it will cause the value to be scheduled to be written to the output buffer. Note that supply lanes should not have backpressure relief so multiple values generated in a single event handler execution need to be queued rather than overwriting. Care should be taken to manage the capacity of the queue.

SirCipher:
Closes #507


SirCipher:
Blocked by #584

server/swim_agent/src/lanes/supply/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+            inner: Default::default(),
+        }
+    }
+
+    fn push(&self, event: T) {
+        let mut guard = self.inner.borrow_mut();
+        guard.event_queue.push_back(event)
+    }
+
+    fn sync(&self, id: Uuid) {
```
horned-sphere, line 79: Is they sync behaviour based on the Java code? What does syncing with a queue even mean...?

server/swim_agent/src/lanes/supply/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+                EventOrSync::Sync => {
+                    if let Some(result) = write(
+                        buffer,
+                        sync_queue,
+                        event_queue,
+                        next,
+                        LaneResponse::<T>::synced,
+                    ) {
+                        return result;
+                    }
```
horned-sphere, line 181: The control flow between this and the write function is very difficult to follow. I think it could be rearranged to be simpler.

horned-sphere (approved):

PR #583, opened 2023-10-17
Adds rust port of the transit app to the examples.

horned-sphere:
Small bug fixes to the runtime for the app.


SirCipher (approved):
Looks really nice!

PR #609, opened 2024-04-04
Outgoing TLS for server

### LINKED ISSUE #604: Add granular support for TLS on the server
Currently `ServerBuilder` can enable TLS support using the `add_tls_support(mut self, config: TlsConfig)` method, however the `TlsConfig` requires both a `ClientConfig` and a `ServerConfig`.

It should be possible to create servers that can open server downlinks with TLS, without requiring them to have TLS support themselves.

DobromirM:
Closes #604 

Changed the rust server to enable outgoing TLS by default for both plain text and TLS servers.

### DIFF
--- a/runtime/swimos_tls/src/net/client.rs
+++ b/runtime/swimos_tls/src/net/client.rs
@@ -29,14 +29,14 @@ use crate::{config::ClientConfig, errors::TlsError, maybe::MaybeTlsStream};
 
 /// [`ClientConnections`] implementation that supports opening both secure and insecure connections.
 #[derive(Clone)]
-pub struct RustTlsClientNetworking {
+pub struct RustlsClientNetworking {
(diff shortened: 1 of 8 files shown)


runtime/swimos_tls/src/net/client.rs
```diff
@@ -29,14 +29,14 @@ use crate::{config::ClientConfig, errors::TlsError, maybe::MaybeTlsStream};
 
 /// [`ClientConnections`] implementation that supports opening both secure and insecure connections.
 #[derive(Clone)]
-pub struct RustlsClientNetworking {
+pub struct RustTlsClientNetworking {
```
SirCipher, line 32: This is actually named after https://crates.io/crates/rustls

SirCipher (approved):

horned-sphere (approved):

PR #611, opened 2024-04-04
Game Demo Port

PiggottDev:
Very rough first draft of the game demo port - needs few iterations of tidying before merging.


example_apps/game/src/agents/model/leaderboard.rs
```diff
@@ -0,0 +1,130 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{cmp::Ordering, sync::Arc};
+
+use super::stats::PlayerTotals;
+
+type Comparator = Arc<dyn Fn(&PlayerTotals, &PlayerTotals) -> Ordering + Sync + Send>;
```
horned-sphere, line 19: Looking at all of the uses of this, you could get rid of this and replace it with `type Comparator = fn(&PlayerTotals, &PlayerTotals) -> Ordering`. There isn't any need for an `Arc`.

Alternatively, could could make this into a trait `trait Comparator: Copy + Send + Sync {... }` and then have `comparator: Box<dyn Comparator>`.

example_apps/game/src/ui.rs
```diff
@@ -0,0 +1,147 @@
...
+        .route("/lib/swim-ui.min.js", get(swim_ui_min_js))
+        .route("/lib/swim-ui.min.js.map", get(swim_ui_min_js_map))
+        .route("/lib/swim-ux.min.js", get(swim_ux_min_js))
+        .route("/lib/swim-ux.min.js.map", get(swim_ux_min_js_map))
+        .route("/lib/swim-vis.min.js", get(swim_vis_min_js))
+        .route("/lib/swim-vis.min.js.map", get(swim_vis_min_js_map))
+        .route("/assets/favicon.ico", get(favicon))
+        .route("/assets/swim-logo.png", get(swim_logo))
+        .with_state(UiConfig { port })
+}
```
horned-sphere, line 50: I realize that this is copied from the example app I did but it's definitely hacky. We shoul probably make somethiung more general if we're going to have a lot of these.

horned-sphere (commented):
Very impressive for a first attempt!

example_apps/game/game-model/src/model/battle.rs
```diff
@@ -0,0 +1,256 @@
...
+            .chain(
+                self.dead
+                    .into_iter()
+                    .map(|player| player.resolve(self.name, is_winner, true))
+            ).collect()
+    } 
+
+}
+
+struct PlayerBattleContext<'a> {
```
SirCipher, line 118: I think this could be better named as something like `PlayerBattleSession|Game|` as could the other things named `..Context`

example_apps/game/game-model/src/model/battle.rs
```diff
@@ -0,0 +1,256 @@
...
+    player: &'a mut Player,
+    kills: usize,
+    assists: usize,
+}
+
+impl<'a> PlayerBattleContext<'a> {
+
+    fn new(player: &mut Player) -> PlayerBattleContext {
```
SirCipher, line 126: A more rusty way of doing this would implementing the `From<&mut Player>` trait for `PlayerBattleContext`

example_apps/game/game-model/src/model/config.rs
```diff
@@ -0,0 +1,95 @@
...
+// TODO: Cache the ThreadRng and Distributions 
+
+pub fn generate_game_duration() -> u64 {
+    rand::thread_rng().sample(Binomial::new(30, 0.5).unwrap()) * 60000 +
+    rand::thread_rng().gen_range(0..60000)
+}
+
+pub fn generate_first_kill_time() -> u64 {
+    rand::thread_rng().sample(Binomial::new(120, 0.3).unwrap()) * 1000 +
+    rand::thread_rng().gen_range(0..1000)
```
SirCipher, line 49: I'd pull these magic numbers out to be consts

example_apps/game/game-model/src/model/game.rs
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+use crate::{player::Player, round::Round};
+use crate::config;
+
+
+#[derive(Debug, Default)]
+pub struct Game {
```
SirCipher, line 21: The name `Game` to me implies a current session and this seems to be the universe's state

example_apps/game/game-model/src/model/game.rs
```diff
@@ -0,0 +1,53 @@
...
+    players: Vec<Player>,
+    match_id: usize,
+}
+
+impl Game {
+
+    pub fn new() -> Game {
+        Game { players: config::generate_all_players(), match_id: 0 }
+    }
```
SirCipher, line 30: A no-arg constructor is typically implemented using the `Default` trait

example_apps/game/game-model/src/model/game.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    pub fn generate_round(&mut self) -> Round {
+
+        let id = self.match_id;
+        self.match_id += 1;
+        if self.match_id == config::MAX_MATCH_ID {
```
SirCipher, line 36: Why the limitation on match ID?

example_apps/game/game-model/src/model/game.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        Round::generate(id, self.generate_round_players())
+    }
+    
+    fn generate_round_players(&mut self) -> Vec<&mut Player> {
+        let ids = config::generate_round_player_ids();
+        // TODO: Has to be a better way of doing this? 
+        // Maybe immutable references with mutable interior state?
+        self.players
```
SirCipher, line 47: It'd be more efficient if you stored the players in a `HashMap` keyed by the player ID so you don't need to iterate every time to generate the round's players

example_apps/game/src/agents/model/leaderboard.rs
SirCipher: The first option would be good here. You could then have a const for each comparator type, something like:
`const TOTAL_KILLS_COMP: for<'a> fn(&'a PlayerTotals, &'a PlayerTotals) -> Ordering = |left, right| left.total_kills.cmp(&right.total_kills);`

example_apps/game/src/agents/model/leaderboard.rs
```diff
@@ -0,0 +1,130 @@
...
+                p1.kd_ratio
+                    .partial_cmp(&p2.kd_ratio)
+                    .unwrap_or_else(|| p1.total_kills.cmp(&p2.total_kills))
+            }),
+        )
+    }
+
+    pub fn xp(capacity: usize) -> Self {
+        Leaderboard::new(capacity, Arc::new(|p1, p2| p1.total_xp.cmp(&p2.total_xp)))
+    }
```
SirCipher, line 58: To tidy this up a bit I would just have one constructor that accepts the comparator

example_apps/game/src/agents/model/leaderboard.rs
```diff
@@ -0,0 +1,130 @@
...
+
+type Comparator = Arc<dyn Fn(&PlayerTotals, &PlayerTotals) -> Ordering + Sync + Send>;
+
+#[derive(Clone)]
+pub struct Leaderboard {
+    comparator: Comparator,
+    // The players ordered by some comparator
+    items: Vec<PlayerTotals>,
+    // The index of each player in the items vector
+    indexes: Vec<Option<usize>>,
```
SirCipher, line 27: I'm not sure I understand this entirely, why not use a `HashMap` and why the `Option<usize>`?

example_apps/game/src/agents/model/leaderboard.rs
```diff
@@ -0,0 +1,130 @@
...
+            self.sort_up(index);
+        }
+    }
+
+    fn sort_up(&mut self, index: usize) -> bool {
+        if index == self.items.len() - 1 {
+            return false;
+        };
+
+        let mover_stats = self.items.get(index).unwrap();
```
SirCipher, line 96: If you're certain that the index exists, you can just get a reference directly using `&self.items[index]`

example_apps/game/src/agents/model/stats.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub start_time: u64,
+    pub end_time: u64,
+    pub first_kill_time: u64,
+    pub stats: MatchStats,
+    pub team_stats: HashMap<String, MatchStats>,
+    pub player_stats: HashMap<usize, MatchStats>,
+}
+
+impl MatchSummary {
+    pub fn from_round(round: Round) -> MatchSummary {
```
SirCipher, line 39: `Impl From<Round> for MatchSummary` would be more ergonomic

example_apps/game/src/agents/model/stats.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+#[derive(Default, Debug, Clone, Form)]
+#[form(fields_convention = "camel")]
+pub struct PlayerTotals {
+    pub id: usize,
+    pub tag: String,
+    pub game_count: usize,
+    // pub total_duration: u64,
+    // pub last_online: u64,
```
SirCipher, line 155: Rogue comment

example_apps/game/src/agents/model/stats.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub loss_count: usize,
+    pub team_count: HashMap<String, usize>,
+}
+
+impl PlayerTotals {
+    pub fn new(id: usize, tag: String) -> Self {
+        Self {
+            id,
+            tag,
+            ..Default::default()
```
SirCipher, line 172: I'm surprised that you found out about this expression, it was a while before I found it!

example_apps/game/src/agents/model/stats.rs
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    pub fn increment(&mut self, match_summary: &MatchSummary) {
+        self.game_count += 1;
+        // self.total_duration += match_summary.duration;
+        // self.last_online = match_summary.end_time;
```
SirCipher, line 179: Rogue comment

example_apps/game/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Bind to a specific port.
+    #[arg(short, long)]
+    port: Option<u16>,
+    /// Bind the UI a specific port.
+    #[arg(short, long)]
+    ui_port: Option<u16>,
+}
+
+fn configure_logging() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let filter = example_filter()?
```
SirCipher, line 93: You should just be able to depend on the `example_util` crate rather than copying this here. You could maybe add a new function there that accepts a slice of additional directives to add

example_apps/game/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+        .await?;
+
+    let (task, handle) = server.run();
+
+    let shutdown = start_agents_and_wait(handle, Some(bound));
+
+    let (_, result) = futures::future::join(shutdown, task).await;
+
+    result?;
+    println!("Server stopped successfully.");
```
SirCipher, line 131: Seeing as the tracing crate is used I'd replace this with an `info!` call

SirCipher (commented):
Looks really great for a first pass! My comments are just nits really. Before this gets merged It could do with a sprinkle of documentation

example_apps/game/game-model/src/model/game.rs
PiggottDev: To re-use match/game agents so it can be left running indefinitely

example_apps/game/game-model/src/model/game.rs
PiggottDev: I use a `Vec` in most places instead of a `HashMap` because the player IDs are `0..n` and so to find a given player ID I can use the index in the `Vec` which should be more efficient than a `HashMap`. The problem I ran into here, and why I iterate over the `Vec`, is that Rust doesn't want me to have two mutable references to the same object so it won't let me select more than one mutable reference by index, it will however let me iterate through them and collect them, I think `HashMap` has the same issue.

example_apps/game/src/agents/model/leaderboard.rs
PiggottDev: This is an attempt to essentially maintain an ordered hash set. The `items: Vec` is the ordered list of players, the `indexes: Vec` is a mapping from player ID to index in the ordered list, `Option` to allow for `None` if the player has not been added to the leaderboard yet.

SirCipher: When you update a `PlayerTotals` I would just remove the existing totals using `Vec::retain`, add the new instance and then call `items.sort_by(|a, b| (self.comparator)(a, b))`

PiggottDev: I think the current implementation is much more efficient, it looks up and removes in constant time whereas `Vec::retain` loops through the list.

SirCipher: Up to you but I think that readability is more important for a demo :) Another option would be a `BTreeMap`

example_apps/game/game-model/src/model/game.rs
horned-sphere: I think this way of doing it is fine for an example (and a `HashMap` wouldb't be that much of an improvement). For what it's worth, I would do the following to select a random selection (here 3 from 5).

```
let mut players = vec![1, 2, 3, 4, 5];
let mut rng = thread_rng();
players.shuffle(&mut rng);
let selected = &mut players[0..3];
```

PR #617, opened 2024-04-25
Resolves incorrect map downlink codec usage.

SirCipher:


swimos_downlink/src/task/map.rs
```diff
@@ -73,7 +73,7 @@ where
             lifecycle,
             FramedWrite::new(output, DownlinkOperationEncoder),
             actions,
-            ValueNotificationDecoder::default(),
+            MapNotificationDecoder::default(),
```
horned-sphere, line 76: I know this is the existing code (and I must have approved it originally!) but I'm failing to see why it being remote or not requires different decoders. Even if it does actually make sense, I think some explanation is needed as a comment.

SirCipher: @horned-sphere, I'm not too sure either. I'll have a look into it

horned-sphere (approved):

PR #620, opened 2024-04-26
Improves combinators that inspect/transform the contents of lanes.

### LINKED ISSUE #606: Add `with_value` handler for value like items in agents.
The `HandlerContext` provides the following handler for map like items:

```rust
    pub fn with_entry<'a, Item, K, V, F>(
        &self,
        lane: fn(&Agent) -> &Item,
        key: K,
        f: F,
    ) -> impl HandlerAction<Agent, Completion = ()> + Send + 'a
    where
        Agent: 'static,
        Item: TransformableMapLikeItem<K, V> + 'static,
        K: Send + Clone + Eq + Hash + 'static,
        V: Clone + 'static, // Clone here is spurious, see #607
        F: FnOnce(Option<V>) -> Option<V> + Send + 'a,
```

This creates a handler that will transform the value associated with a key in the map like item.

This is inflexible as it requires that the value be cloned. It should be altered to borrow the current value instead. An additional combinator called `transform_entry` can be introduced with the existing functionality.

Equivalent combinators should be added for value like items.

horned-sphere:
- Renames the `with_entry` combibnator for map like items to `transform_entry`.
- Adds a new `with_entry` combinator for map like items that computes a value from a borrowed entry in the map.
- Adds `with_value` and `transform_value` combinators for value like items that behave in the same way.

Closes #606.


server/swimos_agent/src/agent_lifecycle/utility/mod.rs
```diff
@@ -201,73 +243,100 @@ impl<Agent: 'static> HandlerContext<Agent> {
...
-        lane: fn(&Agent) -> &Item,
+        item: fn(&Agent) -> &Item,
         key: K,
         f: F,
     ) -> impl HandlerAction<Agent, Completion = ()> + Send + 'a
     where
         Agent: 'static,
-        Item: TransformableMapLikeItem<K, V> + 'static,
+        Item: MutableMapLikeItem<K, V> + 'static,
         K: Send + Clone + Eq + Hash + 'static,
```
SirCipher, line 291: While tangental, I can't see why the Map operations in the `HandlerContext` have a bound of `K: Clone`?

SirCipher (approved):

PR #627, opened 2024-05-21
Tidy up of crate/module API: Stage 1

horned-sphere:
Cleans up the types/functions that are exported from crates and addes crate/module level documentation.

1. Documents the MessagePack crate.
2. Documents the Recon crate.
3. Refactors a lot of components out of `swimos_api` (this had become a bit of dumping ground for things that weren't really part of the public API).
4. Splits the protocol used by the runtime to communicate with agents out (from `swimos_api`) into its own crate (`swimos_agent_protocol`).
5. Documents `swimos_agent_protocol`.


api/swimos_agent_protocol/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub mod ad_hoc {
+        pub use crate::ad_hoc::{
+            AdHocCommandDecoder, AdHocCommandEncoder, RawAdHocCommandDecoder,
+            RawAdHocCommandEncoder,
+        };
+    }
+
+    /// # The protocol used by the runtime to communicate with stores.
+    ///
+    /// TODO Non-transient lanes also implicitly contain a store. They should
```
SirCipher, line 112: Does this need to be part of the doc comment?

api/swimos_agent_protocol/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+            MapLaneRequestDecoder, MapLaneRequestEncoder, MapLaneResponseDecoder,
+            MapLaneResponseEncoder, RawMapLaneRequestDecoder, RawMapLaneRequestEncoder,
+            RawMapLaneResponseDecoder, RawMapLaneResponseEncoder, RawValueLaneRequestDecoder,
+            RawValueLaneRequestEncoder, RawValueLaneResponseDecoder, RawValueLaneResponseEncoder,
+            ValueLaneRequestDecoder, ValueLaneRequestEncoder, ValueLaneResponseDecoder,
+            ValueLaneResponseEncoder,
+        };
+    }
+
+    /// # The encoding used for map like lanes and stores, shared between the other protocols in this module.
```
SirCipher, line 90: This probably doesn't need to be at H1 level

api/swimos_agent_protocol/src/model.rs
```diff
@@ -0,0 +1,212 @@
...
+        key: K,
+    },
+    #[form(tag = "clear")]
+    Clear,
+}
+
+/// Representation of map lane messages (used to form the body of Recon messages when operating)
+/// on downlinks. This extends [`MapOperation`] with `Take` (retain the first `n` items) and `Drop`
+/// (remove the first `n` items). We never use these internally but must support them for communicating
+/// with other implementations.
```
SirCipher, line 79: It might be worth explaining _what_ happens when a take or drop envelope is received

api/swimos_agent_protocol/src/model.rs
```diff
@@ -0,0 +1,212 @@
...
+impl<T> From<StoreResponse<T>> for LaneResponse<T> {
+    fn from(response: StoreResponse<T>) -> Self {
+        let StoreResponse { message } = response;
+        LaneResponse::StandardEvent(message)
+    }
+}
+
+/// Message type for agents to send ad hoc commands to the runtime.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct AdHocCommand<S, T> {
```
SirCipher, line 139: You've renamed the AdHoc encoder/decoder, is it worth renaming this too?

api/swimos_agent_protocol/src/model.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub overwrite_permitted: bool,
+}
+
+impl<S, T> AdHocCommand<S, T> {
+    /// #Arguments
+    /// * `address` - The target lane for the the command.
+    /// * `command` - The body of the command message.
+    /// * `overwrite_permitted` - Controls the behaviour of command handling in the case of back-pressure.
+    /// If this is true, the command maybe be overwritten by a subsequent command to the same target (and so
+    /// will never be sent). If false, the command will be queued instead.
```
SirCipher, line 151: It would be useful to know in what cases this is set to true

api/swimos_agent_protocol/src/lib.rs
horned-sphere: It's like that to make it match with the other modules (that have longer describptions with titles) in this crate, so the formatting is consistent on the doc page. I'm happy to downgrade them though if you think it seems weird?

api/swimos_agent_protocol/src/model.rs
horned-sphere: I'm not sure what you mean (it says that the first `n` messages are dropped or retained in the comment)? I've added longer descriptions to the enum variants to make it a bit clearer.

api/swimos_agent_protocol/src/model.rs
horned-sphere: I'm not sure what you mean. The types are `AdHocCommand`, `AdHocCommandEncoder`, `AdHocCommandDecoder`. That seems consistent to me?

SirCipher: You renamed the decoder here: https://github.com/swimos/swim-rust/pull/627/files#diff-bb2b59058dcb4cc278fca370f6f638e0b9d1486b533753995a9acf3cffbf0551R32

api/swimos_agent_protocol/src/model.rs
SirCipher: It's more that it states that it's not used internally. Perhaps explaining that it will run the command when it's received but it's not possible to issue a take or drop operation internally

horned-sphere: It is possible to do take and drop operations internally, it will just never issue a `Take` or `Drop` Warp frame (I think the Java one does the same these days so, in theory, we'll never receive one either unless someone is using an ancient SwimOS release). I might just remove that comment to avoid confusion.

api/swimos_agent_protocol/src/model.rs
horned-sphere: Ah, that type is supposed to be private. The actual exposed type is still called `AdHocCommandEncoder`. I've removed the spurious `pub`.

SirCipher (approved):

PR #635, opened 2024-06-06
Enable Dependabot

SirCipher:


server/swimos_introspection/Cargo.toml
```diff
@@ -22,7 +22,7 @@ tracing = { workspace = true }
 uuid = { workspace = true }
 thiserror = { workspace = true }
 parking_lot = { workspace = true, features = ["send_guard"] }
-once_cell = "1.17.1"
+once_cell = { workspace = true }
```
horned-sphere, line 25: This should just be remove and replaced with the std lib.

horned-sphere (approved):

PR #655, opened 2024-06-14
Aggregations example application

SirCipher:


example_apps/aggregations/src/aggregate.rs
```diff
@@ -0,0 +1,68 @@
...
+        &self,
+        context: HandlerContext<AggregateAgent>,
+        averages: &HashMap<String, f64>,
+        _key: String,
+        _prev: Option<f64>,
+        _new_value: &f64,
+    ) -> impl EventHandler<AggregateAgent> {
+        let averages = averages.clone();
+        context
+            .effect(move || averages.values().sum::<f64>() / averages.len() as f64)
```
horned-sphere, line 50: There's no reason to put this into an effect. You can just compute the average without doing the clone and then set it into the lane.

horned-sphere: Also, if you wanted to be particularly efficient, you update the average rather than recompute it each time.

example_apps/aggregations/src/area.rs
```diff
@@ -0,0 +1,134 @@
...
+    agent::event_handler::HandlerActionExt,
+    agent::lanes::{CommandLane, JoinValueLane, ValueLane},
+    agent::lifecycle,
+    agent::projections,
+    agent::AgentLaneModel,
+};
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum Area {
+    A,
```
horned-sphere, line 33: Should probably give them (possibly made up but plausible) place names. This looks a bit cryptic.

example_apps/aggregations/src/area.rs
```diff
@@ -0,0 +1,134 @@
...
+            "B" => Ok(Area::B),
+            "C" => Ok(Area::C),
+            "D" => Ok(Area::D),
+            _ => Err(()),
+        }
+    }
+}
+
+impl Area {
+    pub fn random() -> Self {
```
horned-sphere, line 54: This might be better as something like `select_random` to make it clearer.

example_apps/aggregations/src/area.rs
```diff
@@ -0,0 +1,134 @@
...
+            .copied()
+            .expect("Slice was not empty")
+    }
+}
+
+#[derive(AgentLaneModel)]
+#[projections]
+pub struct AreaAgent {
+    register: CommandLane<u64>,
+    deregister: CommandLane<u64>,
```
horned-sphere, line 66: This could be one lane with a custom message type.

example_apps/aggregations/src/area.rs
```diff
@@ -0,0 +1,134 @@
...
+    cars: JoinValueLane<u64, u64>,
+    average_speed: ValueLane<f64>,
+}
+
+#[derive(Clone, Default)]
+pub struct AreaLifecycle {
+    area: Arc<Mutex<Option<Area>>>,
```
horned-sphere, line 73: This could be a `Cell` as `Area` is copy. Even if that weren't the case, `RefCell` would have been good enough. I went to huge trouble to make it so you don't need mutexes in lifecycles (unless they need to be mutated in futures)! Alternatively, as the set of areas is finite, you could put this into a constructor (see the transit app).

horned-sphere: It might be nice to be able to give the `LifecycleFactory` access to the parameter map, in general, so that we can make this sort of intiailization easier.

example_apps/aggregations/src/area.rs
```diff
@@ -0,0 +1,134 @@
...
+        &self,
+        context: HandlerContext<AreaAgent>,
+        speeds: &HashMap<u64, u64>,
+        _key: u64,
+        _prev: Option<u64>,
+        _new_value: &u64,
+    ) -> impl EventHandler<AreaAgent> {
+        let speeds = speeds.clone();
+        context
+            .effect(move || speeds.values().sum::<u64>() as f64 / speeds.len() as f64)
```
horned-sphere, line 131: Another unnecessary clone/effect.

example_apps/aggregations/src/car.rs
```diff
@@ -0,0 +1,101 @@
...
+
+#[derive(AgentLaneModel)]
+#[projections]
+pub struct CarAgent {
+    speed: ValueLane<u64>,
+}
+
+#[derive(Debug, Clone)]
+pub struct CarLifecycle {
+    area: Arc<Mutex<Area>>,
```
horned-sphere, line 39: Given that this is mutable, I'd stick in in a transient `ValueStore`.

horned-sphere (approved):

PR #660, opened 2024-06-20
Data windowing example application

SirCipher:


example_apps/time_series/src/agent.rs
```diff
@@ -0,0 +1,158 @@
...
+        &self,
+        context: HandlerContext<ExampleAgent>,
+        key: u64,
+    ) -> impl EventHandler<ExampleAgent> {
+        match self {
+            RetentionPolicy::Count { max, keys } => {
+                let timestamps = &mut *keys.borrow_mut();
+                timestamps.push_front(key);
+
+                let drop = timestamps.len().checked_sub(*max).unwrap_or_default();
```
horned-sphere, line 112: I know it's correct but I feel putting the `if` out in the code would make the intention clearer. (Also, I probbaly have done this myself somewhere, so sorry...)

example_apps/time_series/src/agent.rs
```diff
@@ -0,0 +1,158 @@
...
+                    let keys = timestamps
+                        .split_off(drop)
+                        .into_iter()
+                        .take(drop)
+                        .map(move |key| context.remove(ExampleAgent::HISTORY, key));
+                    Some(Sequentially::new(keys))
+                } else {
+                    None
+                };
+                handler.discard().boxed()
```
horned-sphere, line 123: You could probably rejig this to not require boxing but it's probably not worth it.

horned-sphere (approved):

PR #669, opened 2024-06-24
Code coverage for CI

DobromirM:


.tarpaulin.toml
```diff
@@ -1,43 +1,9 @@
 [coverage]
-exclude = [
```
SirCipher, line 2: Could you add these back in? They negatively affect the coverage report

SirCipher (approved):

PR #670, opened 2024-06-24
Sever client networking cryptographic provider initialisation

SirCipher:


horned-sphere (commented):
I think it would be better to allow the user to specify the default `CryptoProvider` rather than adding it to the configuration. As we're providing a library, the Swim server could form part of a larger application where the user is relying on setting the default crypto provider an they want it used everywhere.

We should then either add a builder method to the server, or a feature flag to the crate (probably enabled by default) that will set the `CryptoProvider` to _our choice_ that can be used by people who don't want to care about it.

runtime/swimos_remote/src/tls/mod.rs
```diff
@@ -23,3 +23,40 @@ pub use config::{
...
+    #[default]
+    FromFeatureFlags,
+    Provided(Arc<CryptoProvider>),
+}
+
+impl CryptoProviderConfig {
+    pub fn build(self) -> Arc<CryptoProvider> {
+        match self {
+            CryptoProviderConfig::ProcessDefault => CryptoProvider::get_default()
+                .expect("No default cryptographic provider specified")
```
horned-sphere, line 41: This should be an error rather than a panic.

runtime/swimos_remote/src/tls/errors.rs
```diff
@@ -32,4 +32,6 @@ pub enum TlsError {
     /// Performing the TLS handshake failed.
     #[error("TLS handshake failed: {0}")]
     HandshakeFailed(std::io::Error),
+    #[error("Invalid cryptographic provider configured: {0}")]
+    InvalidCryptoProvider(String),
```
horned-sphere, line 36: Missing doc comment.

horned-sphere: Also, there is only ever one string this can be so you may as well remove the `String` field and just put that string into the `#error` tag.

horned-sphere (approved):

PR #676, opened 2024-06-27
Adds missing cargo flags

SirCipher:
Adds missing cargo versions and features following crate releases

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,13 @@ members = [
     "example_apps/devguide/2_3/*",
 ]
 
+[workspace.package]
+version = "0.1.0"
+authors = ["Swim Inc. developers info@swim.ai"]
+edition = "2021"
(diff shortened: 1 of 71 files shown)


horned-sphere (commented):
It's possible to move all of this into the workspace Cargo.toml.

For example, for swimos_api, you can put:

```swimos_api = { path = "api/swimos_api", version = "0.1.0" }```

And then changes the deps to:

```swimos_api = { workspace = true }```

(I just tried it and it works).

(Also, maybe could use https://doc.rust-lang.org/nightly/cargo/reference/workspaces.html#the-package-table to put the version number in one place?)

SirCipher:
@horned-sphere, that's not a bad idea. It's actually better supported in IDEs now too as they used to shout at me for doing it.

horned-sphere (approved):

PR #680, opened 2024-06-28
Improves client downlink lifecycle type inference

SirCipher:


swimos_client/src/lib.rs
```diff
@@ -373,7 +373,11 @@ pub struct ValueDownlinkBuilder<'h, L> {
 
 impl<'h, L> ValueDownlinkBuilder<'h, L> {
     /// Sets a new lifecycle that to be used.
-    pub fn lifecycle<NL>(self, lifecycle: NL) -> ValueDownlinkBuilder<'h, NL> {
+    pub fn lifecycle<NL, T>(self, lifecycle: NL) -> ValueDownlinkBuilder<'h, NL>
+    where
+        L: ValueDownlinkLifecycle<T>,
```
horned-sphere, line 378: I don't think the he `L` bound here isn't really needed (and should go on the impl block if it is).

horned-sphere (approved):

PR #686, opened 2024-07-10
Client examples for server applications

### LINKED ISSUE #679: Example applications missing client binaries
The example applications should have client cargo examples

SirCipher:
Closes #679


example_apps/aggregations/examples/aggregations_client.rs
```diff
@@ -0,0 +1,40 @@
...
+        .on_event_blocking(|speed| println!("/cars/1 speed: {speed}"));
+    let _car_view = client_handle
+        .value_downlink::<u64>(RemotePath::new(host, "/cars/1", "speed"))
+        .lifecycle(car_lifecycle)
+        .open()
+        .await?;
+
+    let area_lifecycle = BasicValueDownlinkLifecycle::default()
+        .on_event_blocking(|speed| println!("/area/arbury average speed: {speed}"));
+    let _area_view = client_handle
```
horned-sphere, line 23: This feels a bit odd for an example. Is it necessary to hold onto the views to keep the downlinks open (even though you might not need to use them). If so, it feels like the API could be improved.

horned-sphere: Maybe an `open_detached` method?

example_apps/join_map/src/agents.rs
```diff
@@ -34,6 +34,8 @@ impl AggregatedLifecycle {
         &self,
         context: HandlerContext<AggregatedStatisticsAgent>,
     ) -> impl EventHandler<AggregatedStatisticsAgent> {
+        println!("Start agg");
```
horned-sphere, line 37: Rouge print statement.

example_apps/join_value/Cargo.toml
```diff
@@ -6,5 +6,12 @@ publish = false
 
 [dependencies]
 swimos = { workspace = true, features = ["server", "agent"] }
+swimos_client = { workspace = true }
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 example-util = { path = "../example_util" }
+
+[dev-dependencies]
+rand.workspace = true
```
horned-sphere, line 14: Why does an example need a dev dependency?

runtime/swimos_runtime/src/downlink/mod.rs
```diff
@@ -510,10 +511,10 @@ impl<D: Decoder> DownlinkReceiver<D> {
     }
 }
 
-struct Failed(u64);
+struct Failed<E>(u64, E);
```
horned-sphere, line 514: Having a generic error type rather than usin `dyn` feels a bit odd.

swimos_client/src/transport.rs
```diff
@@ -189,6 +189,7 @@ where
 
             match event {
                 TransportEvent::Request(TransportRequest::Resolve(shp, callback)) => {
+                    trace!(%shp, "Resolving host");
```
horned-sphere, line 192: `shp` is an unhelpful field name for a log message. ⛵

swimos_client/src/transport.rs
```diff
@@ -205,14 +206,17 @@ where
                     addrs,
                     callback,
                 }) => {
+                    trace!(?host, ?scheme, "Resolving connection");
```
horned-sphere, line 209: Some of these log messages are a bit vague in isolation.

swimos_client/src/transport.rs
```diff
@@ -205,14 +206,17 @@ where
...
                     let peer = addrs
                         .iter()
                         .find_map(|sock| peers.get(sock).map(|tx| (sock, tx)));
                     match peer {
                         Some((sock, peer)) => {
+                            trace!("Found peer");
```
horned-sphere, line 215: Should add some fields where the are available.

horned-sphere (commented):
You've made the clients to be examples in each of the projects. I feel like they should be alternative bin targets instead. They're not really examples.

swimos_client/src/transport.rs
SirCipher: These were leftovers from some debugging so I'll remove them actually

example_apps/join_value/Cargo.toml
SirCipher: That's where dependencies go for example binaries. It's used to generate data in the client

example_apps/aggregations/examples/aggregations_client.rs
SirCipher: It's not necessary to keep the handle. Dropping it will cause the runtime to switch to a read-only mode which fires the event handlers. I'll remove the binding as it is a bit misleading

horned-sphere (approved):

PR #689, opened 2024-07-12
Adds Hickory DNS support back into the server

### LINKED ISSUE #647: Trust DNS resolver no longer configured
The Trust DNS resolver is no longer setup correctly using cargo features as the argument is not passed to the swimos_remote crate.

SirCipher:
Closes #647


horned-sphere (commented):
trust-dns seems to have rebranded to [this](https://github.com/hickory-dns/hickory-dns). Probably should switch.

horned-sphere (approved):

PR #691, opened 2024-07-18
Ripple example application

### LINKED ISSUE #589: Port ripple demo

SirCipher:
Closes #589


example_apps/ripple/src/agent.rs
```diff
@@ -0,0 +1,398 @@
...
+        /// The ID of the charge to remove.
+        id: Id,
+    },
+}
+
+/// The agent's lifecycle.
+#[derive(Clone)]
+pub struct MirrorLifecycle {
+    /// The random number generator to use.
+    rng: Arc<Mutex<StdRng>>,
```
horned-sphere, line 168: Could you not just use a thread local RNG?

example_apps/ripple/src/util.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::agent::Color;
+use std::fmt::Display;
+use swimos::model::{Text, ValueKind};
+use swimos_form::read::{
+    ExpectedEvent, ReadError, ReadEvent, Recognizer, RecognizerReadable, SimpleAttrBody,
+    SimpleRecBody,
+};
+use swimos_form::write::{PrimitiveWriter, StructuralWritable, StructuralWriter};
+
+impl RecognizerReadable for Color {
```
horned-sphere, line 24: Implementing traits for type in a different module is a bit weird. (I could see the recognizers being shuffled out of the way). Also, hand implemented forms in an example is a bit icky (but I guess you can't avoid that if you're reusing the UI code).

horned-sphere: So, looking again, the issue is the UI can specify an arbitray string for the colour but actually only ever uses those 3 options (which you need to be able to choose between randonly)? I think it might be better to leave the field in the type as a string so you don't need the custom form. You can still choose between three specific colours internally without it being an enum. I think it's best to avoid custom forms in examples (unless that is the point of the example).

SirCipher: Yeah, that's fair. It just seemed ugly but I guess this isn't much better either :)

example_apps/ripple/src/agent.rs
SirCipher: I probably could but also wanted to demonstrate lifecycles

horned-sphere: I don't really think we should be encouraging mutexes (they should generalyl be a last resort). Also, I think in this case, I think it would even be possible for the RNG to belong to the stream itself and not need to be shared at all. (Regardless, I also notice you've typedefed this above but then not used it here).

horned-sphere (approved):

PR #693, opened 2024-07-19
Simulated Stocks Demo

SirCipher:


example_apps/stocks_simulated/src/agent.rs
```diff
@@ -0,0 +1,160 @@
...
+            let mut rng = ThreadRng::default();
+            let handler = generate_stock(context);
+            let next_delay = Duration::from_secs(rng.gen_range(5..=10));
+
+            Some((handler, next_delay))
+        });
+        context
+            .get_parameter("symbol")
+            .and_then(move |symbol_opt: Option<String>| {
+                let symbol = symbol_opt.expect("Missing symbol for stock");
```
horned-sphere, line 141: I've added a mechanism to kick errors out to the runtime in the connector branch, not that it helps you here.

horned-sphere (approved):

example_apps/stocks_simulated/src/agent.rs
SirCipher: That would be helpful. It'd also be helpful to have access to all parameters in the `on_start` function

PR #697, opened 2024-08-08
Adds the first version of the connector API and Kafka connector.

horned-sphere:
Adds an API for writing ingress connectors.

Implements the connector API for Apache Kafka as an example.

Top level docs are missing pending the implementation of a few more connectors which might alter the API a bit.


server/swimos_connector_kafka/src/selector/mod.rs
```diff
@@ -0,0 +1,690 @@
...
+impl From<MessagePart> for MessageField {
+    fn from(value: MessagePart) -> Self {
+        match value {
+            MessagePart::Key => MessageField::Key,
+            MessagePart::Payload => MessageField::Payload,
+        }
+    }
+}
+
+/// A lazy loader fro a component of a Kafka messages. This ensures that deserializers are only run if a selector
```
SirCipher, line 52:
```suggestion
/// A lazy loader for a component of a Kafka messages. This ensures that deserializers are only run if a selector
```

server/swimos_connector_kafka/src/selector/mod.rs
```diff
@@ -0,0 +1,690 @@
...
+where
+    F: Fn() -> Result<Value, DeserializationError>,
+{
+    fn get(&mut self) -> Result<&Value, DeserializationError> {
+        let Computed { inner, f } = self;
+        if let Some(v) = inner {
+            Ok(v)
+        } else {
+            *inner = Some(f()?);
+            Ok(inner.as_ref().expect("Should be defined."))
```
SirCipher, line 91:
```suggestion
            Ok(inner.insert(f()?))
```

SirCipher (approved):

server/swimos_agent/src/agent_model/mod.rs
```diff
@@ -718,13 +747,52 @@ where
...
+                            DynamicRegistrationError::DuplicateName(name),
+                        )));
+                        dyn_lane_handlers.push(handler);
+                    }
+                }
+            };
+        }
+
+        handle_dyn_lanes!();
+
```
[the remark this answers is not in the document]
horned-sphere: I agree it's fairly monstrous. The issue is that the future generated by the function needs to be `Send` which means that no references to non-`Send` things can be held across awaits. That, in combination with unnamable types make it very hard to break out sub-functions (I've tried, _a lot_). The macros are the best I could come up with to improve the readability.

PR #712, opened 2024-09-27
Reject non-warp0 upgrade requests

SirCipher:


runtime/swimos_http/src/websocket.rs
```diff
@@ -96,7 +95,7 @@ where
 }
 
 /// Produce a bad request response for a bad websocket upgrade request.
-pub fn fail_upgrade(error: ratchet::Error) -> Response<Full<Bytes>> {
+pub fn fail_upgrade(error: impl ToString) -> Response<Full<Bytes>> {
```
horned-sphere, line 98: Maybe `: Display` makes more sense?

horned-sphere (approved):

PR #730, opened 2024-10-11
Adds a future that consumes the task budget.

horned-sphere:


SirCipher:
Actually, maybe not as it's an async fn and doesn't return Poll.

SirCipher (approved):

PR #741, opened 2024-11-27
Improves deserialization efficiency for agents.

horned-sphere:
Agent implementations can now keep persistent deserializers for their lanes rather than recreating them each time.

The agents created by the derive macro have been updated to make use of this.

### DIFF
--- a/server/swimos_agent_derive/src/lane_model_derive/mod.rs
+++ b/server/swimos_agent_derive/src/lane_model_derive/mod.rs
@@ -214,13 +214,23 @@ impl<'a> ToTokens for DeriveAgentLaneModel<'a> {
 
             #[automatically_derived]
             impl #root::agent_model::AgentSpec for #agent_type {
-                type ValCommandHandler = #value_handler;
+                type ValCommandHandler<'a> = #value_handler
+                where
+                    Self: 'a;
 
-                type MapCommandHandler = #map_handler;
+                type MapCommandHandler<'a> = #map_handler
+                where
+                    Self: 'a;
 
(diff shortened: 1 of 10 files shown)


server/swimos_connector/src/generic/mod.rs
```diff
@@ -174,22 +174,45 @@ impl AgentDescription for ConnectorAgent {
     }
 }
 
+/// A single deserializer for [`Value`] objects, shared between all lanes of a [`ConnectorAgent`].
+#[derive(Default)]
+pub struct GenericDeserializer {
```
SirCipher, line 179: Generic doesn't quite feel right for this as it implies `<T>` to me instead of `Value`

server/swimos_agent_derive/src/lane_model_derive/mod.rs
```diff
@@ -545,6 +574,34 @@ impl<'a> HttpHandlerType<'a> {
...
+            }
+            WarpLaneSpec::Map(k, v, _) => {
+                quote!((#root::ReconDecoder<#k>, #root::ReconDecoder<#v>))
+            }
+            WarpLaneSpec::Demand(_)
+            | WarpLaneSpec::DemandMap(_, _)
+            | WarpLaneSpec::JoinValue(_, _)
+            | WarpLaneSpec::JoinMap(_, _, _)
+            | WarpLaneSpec::Supply(_) => {
+                quote!(())
```
SirCipher, line 599: This branch could do with a comment explaining why it's `()`

